PR #13, opened 2013-07-17
Additional compatibility fixes for Robot Framework 2.8+

userzimmermann:


userzimmermann:
Hi. It's been a while. As I've seen that RIDE development continues, is there any chance that these fixes get merged now? They are still working with latest Robot releases, even 2.9a1. And is there a chance that future official RIDE releases don't include a bundled Robot anymore? It would just be nice if we would no longer need to maintain a custom RIDE distribution at work. It just happens again and again that people install official RIDE by mistake and then our Libraries and Tests are not working  and the people don't know why and the mail and phone terror starts... ;)

yanne:
Yeah, been a while, it seems.

I made a new branch, https://github.com/robotframework/RIDE/tree/RF2.8-compatibility, which incorporates your changes, adds a couple of more fixes and updates the bundled RF to 2.8.7.

I plan to make 1.4.alpha-1 based on that branch.

As for dropping the bundled RF, that's somewhat more involved, since Python does not allow to install multiple versions of the same package very easily. There will probably always be lag between releasing of RF version and compatible RIDE version. If RF is not bundled with RIDE, it may mean that people can't upgrade RF, because that would break RIDE. This is a situation we want to avoid, and the current solution is to bundle a compatible RF with RIDE. 

One option would of course to just copy-paste the RF code we need inside RIDE, since that's mostly parsing logic and it changes relatively rarely. That just seems wrong, though.

This PR can be closed though.

yanne:
Ah, one more thing. @userzimmermann: can you test my branch briefly and see if it works for you? That would be a fine validation to have, even before releasing any alpha version.

userzimmermann:
@yanne Thanks :) I will test it.
Regarding the RF bundle: I would just define RF as a requirement with version constraints. Then check the RF version on starting RIDE again and tell the user that they have an incompatible RF version, what to do to install a compatible version and that experiments with newer RF versions should be done in a separate Python environment (what most people will do anyway when testing new stuff).

userzimmermann:
...and add a command line option for disabling the RF version check :)

HelioGuilherme66:
Sorry for asking, but did you miss to push to remote? I can't find the
branch :(.

yanne:
Nope, I was just too fast. It's already merged (and remote branch removed). It's also included in 1.4 alpha 1 that was released today.

drobati:
:+1: Thank you for the hard work.

jussimalinen:
@userzimmermann How installing a ride with bundled robot causes your tests to fail? The whole point of bundling the robot with ride was to make Ride independent of test running. The test runner does not use the bundled robot, but the installed one. And the the bundled robot should not be seen by any tests nor libraries?

userzimmermann:
@jussimalinen I should have made this clearer. Running the tests works of course. The problem is the editing. Most of our libraries depend on my robotframework-tools package, which depends on RF 2.8+. So the complains are more like "Why are all the imports red?", "Why are the Keywords not recognized?", "It doesn't find any of the Keywords!", "Why don't I get any Keyword info?", ... :)

PR #15, opened 2013-08-04
Complete command to launch ride with 32 bit architecture

cdmatta:


yanne:
I made some changes myself, and will include this commit in another branch

PR #26, opened 2015-04-02
path name with "&" should be supported

yangyzho:
it may report an error as below,
this is because of the robotframework_ride not handle the special
characters "&" when it  format the  command, one way of handle this
exception is modify the file:
C:\Python27\Lib\site-packages\robotframework_ride-x.x.x-pyx.x.egg\robotide\contrib\testrunner\
testrunnerplugin.py .


yanne:
Can you modify the commit message so that it's easier to read what this commit fixes.

Something along the lines of `Support & in path names in test runner plugin`

You can also add yourself to the `AUTHORS` file

yanne:
Created #28 with better commit message but the same content

PR #31, opened 2015-06-02
Set min height to tag box

yanne:
This is required in OSX, otherwise the tag box height is calculated wrong
initially.


yanne:
Fixes https://code.google.com/p/robotframework-ride/issues/detail?id=802

Issue #1469, opened 2015-06-15
iCCP: known incorrect sRGB profile


Twoflower2:
"iCCP: known incorrect sRGB profile" error in openSuse KDE 4.14.9 with plasma 5 using RIDE = 1.4b3 or 1.3

yanne:
Is this fatal, i.e. does it prevent RIDE from starting at all?

yanne:
Fixed in ad48e3be6eb97b1ae4b8ee01537770bbe6c07e77

Issue #924, opened 2015-06-02
Support wxPython 3.0


yanne:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=888) by @yanne on 17 Nov 2011_

We explicitly select wxPython 2.8 if it is found using  wxversion.select('2.8') in roboide/**init**.py

However, wxPython 2.9 has been available for some time, so we should prefer that if it's found.

yanne:
> *Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=888#c1) by jerry57 on 17 Nov 2011*


On a Lion laptop I installed wxPython 2.9.2.4 cocoa and then removed the ride install that was there and installed 0.39.1.  When I tried to launch ride.py I am getting the following.

Note: previous 2.8.12.1-1 wxPython was uninstalled and env for VERSIONER
(unset VERSIONER_PYTHON_PREFER_32_BIT)
was removed.

IPC-Freud:de jerry$ ride.py . &amp;
[1] 645
IPC-Freud:de jerry$ Traceback (most recent call last):
  File "/usr/local/bin/ride.py", line 22, in &lt;module&gt;
    main(sys.argv[1:])
  File "/Library/Python/2.7/site-packages/robotide/__init__.py", line 59, in main
    _run(*args)
  File "/Library/Python/2.7/site-packages/robotide/__init__.py", line 68, in _run
    ride = RIDE(inpath)
  File "/Library/Python/2.7/site-packages/robotide/application/application.py", line 35, in __init__
    wx.App.__init__(self, redirect=False)
  File "/usr/local/lib/wxPython-2.9.2.4/lib/python2.7/site-packages/wx-2.9.2-osx_cocoa/wx/_core.py", line 8593, in __init__
    self._BootstrapApp()
  File "/usr/local/lib/wxPython-2.9.2.4/lib/python2.7/site-packages/wx-2.9.2-osx_cocoa/wx/_core.py", line 8158, in _BootstrapApp
    return _core_.PyApp__BootstrapApp(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/robotide/application/application.py", line 40, in OnInit
    self.frame = RideFrame(self, self._controller)
  File "/Library/Python/2.7/site-packages/robotide/ui/mainframe.py", line 63, in __init__
    self._init_ui()
[...]

PR #1479, opened 2015-07-19
Fix the "iCCP: known incorrect sRGB profile" libpng error which was popping when these icons were used.

goodwillcoding:
All the pngs were scanned using:
while read file; do _test=$(identify -verbose $file | grep -i "png:icc"); if [ $? -eq 0 ]; then echo "$file: $_test"; fi  done< <(find . -name "*.png")

and fixed using: convert <file> -strip <file>

this should address issue #1469 and also comments in
https://github.com/robotframework/RIDE/issues/924#issuecomment-107946789

The fix strip the iCCP profile fromt he 5 images what


yanne:
Thanks for the patch!

Issue #1473, opened 2015-06-24
Version 1.4b3 forces *.robot file selection on Windows even though Default file format=txt


VipSaran:
A suggestion for convenience:

The `Default file format` preference value should be a pointer to which filter to apply to Open file dialog (on Windows), so that if "txt" is choosen, then this filter should be pre-selected:
![2015-06-24_135550](https://cloud.githubusercontent.com/assets/995948/8329325/d14614aa-1a78-11e5-8c95-d5ec5a16b400.png)

Similarily, if the `Default file format` value is "robot", then the filter should be:
![2015-06-24_113147](https://cloud.githubusercontent.com/assets/995948/8329343/f8ccaa2a-1a78-11e5-99b4-dd913cd4707e.png)

And for other  `Default file format` preference values (tsv,html), the filter should be "All":
![2015-06-24_135827](https://cloud.githubusercontent.com/assets/995948/8329356/1d71b7ee-1a79-11e5-9d2d-7ac2b75834f8.png)

Why?
As all my robot fw files are *.txt, I have to each time manually change the filter in the Open file dialog.

So, if not change it to be based on `Default file format` preference value, at least make it remember the last selected filter. Thanks.

yanne:
This is fixed by #1482

VipSaran:
Works as expected :+1:

PR #1482, opened 2015-07-20
Use default file format as default wildcard

yanne:
Also removed the duplication between opening suites and opening resource
files


yanne:
Fixes #1473

Issue #1484, opened 2015-07-21
PyDeadObjectError is thrown when preferences dialog is closed in Windows


yanne:
This traceback can be seen in the cmd window:

`File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 14669, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\lib\masked\numctrl.py", line 1171, in _CheckInsertionPoint
    sel_start, sel_to = self._GetSelection()
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 14619, in __getattr__
    raise PyDeadObjectError(self.attrStr % self._name)
wx._core.PyDeadObjectError: The C++ part of the NumCtrl object has been deleted,
 attribute access no longer allowed.`

PR #1485, opened 2015-07-21
Fix errors in preferences dialog closing on Windows

yanne:
For some reason, closing the preferences dialog causes a number of
tree selection events to be triggered. Ignore those, since handling
those might cause PyDeadObjectErrors


yanne:
Fixes #1484

Issue #1476, opened 2015-06-29
Standard library imports may fail if a different major version of RF is installed 


VipSaran:
RIDE (v1.4) shows a proper "hint" when it doesn't recognize a keyword:
![2015-06-29_105440](https://cloud.githubusercontent.com/assets/995948/8404687/411ec7f2-1e4f-11e5-80be-0137dc2c9669.png)

But, since this is a Robot built-in keyword:
![2015-06-29_110656](https://cloud.githubusercontent.com/assets/995948/8404712/75294234-1e4f-11e5-91ce-7365644be72c.png)

it should be automatically recognized (w/o any manual linking/importing), as it was in v1.1.

VipSaran:
The error doesn't appear on RF 2.8.7.
The version of RF I reported the issue is 2.7.2, so this seems to be some sort of a problem with backwards compatibility.

yanne:
Actually the problem is more deep than that.

It seems that the fix to #1371 is quite seriously flawed. This causes some standard libraries not being correctly imported, if a different major version of RF is installed.

More technically, I tried to fix #1371 so that only the `robot/libraries` from the system RF installation is put into the PYTHONPATH.  This works, but *if* those libraries require other `robot` modules, they are imported from the bundled RF, not the system installation. If there are API changes in those modules, the import will fail, causing the library to be not correctly recognised.

PR #1487, opened 2015-07-24
Fix Remote imports with system RF libs

yanne:
When using installed RF libraries, it is still required to have
`Remote` in PYTHONPATH, because remote library keywords cannot
be known statically. The easiest way to do this is to have `Remote.py`
put separately in the PYTHONPATH.


yanne:
Related to #1476

Issue #1500, opened 2015-08-13
'--monitorcolors' and '--monitorwidth' is deprecated WARN message


jongu82:
Running a test in RIDE using the latest release version produces the following warning message in the console:

[ WARN ] Option '--monitorcolors' is deprecated. Use '--consolecolors' instead.
[ WARN ] Option '--monitorwidth' is deprecated. Use '--consolewidth' instead.

According to @yanne:
Those options were changed in RF 2.9, but RIDE run plugin uses the old ones.

PR #1505, opened 2015-08-21
#1500: Remove deprecated options warning messages

timjohncurtis:
-Moved all instances of '--monitorcolors' to '--consolecolors'
-Moved all instances of '--monitorwidth' to '--consolewidth'

'invoke test' passes with no issues reported

### DIFF
--- a/src/robotide/contrib/testrunner/testrunner.py
+++ b/src/robotide/contrib/testrunner/testrunner.py
@@ -171,7 +171,7 @@ class TestRunner(object):
         self._process = Process(cwd)
         self._process.run_command(command)
 
-    def get_command(self, profile, pythonpath, monitor_width, names_to_run):
+    def get_command(self, profile, pythonpath, console_width, names_to_run):
         '''Return the command (as a list) used to run the test'''
         command = profile.get_command_prefix()[:]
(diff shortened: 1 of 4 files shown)


pekkaklarck:
The problem with this change is that the new command line options are not compatible with older Robot Framework versions. I don't think not supporting RF 2.8.7 and earlier as a runner is an option. A workaround is using the short options `-C` and `-W` that have not changed.

I consider it a mistake we deprecated the old options in RF 2.9 when we introduced the new ones. We could consider removing the deprecation in RF 2.9.1, adding it back in RF 3.0, and finally removing it in RF 3.1. There might be needs to keep using the RF versions even then, though, so perhaps using `-C` and `-W` is the best solution.

timjohncurtis:
I have updated the branch to just use the short options, which are supported in both versions, I have tested this with robot framework 2.9 and 2.8

I have added comments where the short options are put on their own so it is not ambiguous, maybe the help text in ./src/robotide/lib/robot/rebot.py should also be updated as well, to include both the 2.9 version and the 2.8 version, then raise another ticket in backlog to remove them when 2.8 support is removed from RIDE.

yanne:
@timjohncurtis, thanks for the PR.

PR #1517, opened 2015-09-04
Updates arguments for pybot to RF 2.9.1.

HelioGuilherme66:
Removes the error "option ... not found" and the red color in runner Arguments.


yanne:
Thanks for the patch!

Issue #1515, opened 2015-09-03
Content Assist fails with unicode characters


serhatbolsu:
Hello,

I am using turkish named written keywords and variables such as;
&{mainpage_Hesabım_Login_ŞifremiUnuttum}

As a result:
--------------------
ciceksepetis-iMac:~ test$ ride
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 49, in OnChar
    self.show_content_assist()
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 100, in show_content_assist
    if self._populate_content_assist():
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 116, in _populate_content_assist
    return self._popup.content_assist_for(value, row=self._row)
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 261, in content_assist_for
    self._choices = self._suggestions.get_for(value, row=row)
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 200, in get_for
    self._previous_choices = self._get_choices(value, row)
  File "/Library/Python/2.7/site-packages/robotide/editor/contentassist.py", line 214, in _get_choices
    choices = self._suggestion_source.get_suggestions(value, row)
  File "/Library/Python/2.7/site-packages/robotide/namespace/suggesters.py", line 26, in get_suggestions
    return self._controller.get_local_namespace_for_row(row).get_suggestions(value)
  File "/Library/Python/2.7/site-packages/robotide/namespace/local_namespace.py", line 48, in get_suggestions
    suggestions = LocalMacroNamespace.get_suggestions(self, start)
  File "/Library/Python/2.7/site-packages/robotide/namespace/local_namespace.py", line 32, in get_suggestions
    return self._namespace.get_suggestions_for(self._controller, start)
  File "/Library/Python/2.7/site-packages/robotide/namespace/namespace.py", line 138, in get_suggestions_for
    sugs.update(self._variable_suggestions(controller, start, ctx))
  File "/Library/Python/2.7/site-packages/robotide/namespace/namespace.py", line 166, in <genexpr>
    sugs = (v for v in variables if v.name_matches(start))
  File "/Library/Python/2.7/site-packages/robotide/namespace/namespace.py", line 363, in __iter__
    name = '{0}{{{1}}}'.format(prefix, name)
[...]

PR #1521, opened 2015-09-14
Fix UnicodeEncodeErrors in `str.format()` usages

yanne:
It seems that str.format() breaks with UnicodeEncodeError if
str is ascii literal and format arguments are unciode. Fix all usages
of `format` to prevent runtime errors


yanne:
Fixes #1515

PR #1563, opened 2016-01-15
Make test runner console font less ugly

GLMeece:
Changed font designation from SYS_SYSTEM_FIXED_FONT to SYS_ANSI_FIXED_FONT. SYS_SYSTEM_FIXED_FONT creates a truly bugly experience. SYS_ANSI_FIXED_FONT isn't perfect, but it's a bit more tolerable.


yanne:
Well, I can't see a reason why not, the new font is easier to the eyes.

Issue #1527, opened 2015-09-22
Tag list sort by name in View all tags.


synsun:
In "View all tags", all tags are sorted by occurrences.

I want see this list sorted by name if I clicked "Tag" title bar.

It could help you find similar tags in a see.

Thanks.

yanne:
Fixed in #1565

PR #1565, opened 2016-01-17
Enhancement #1527 sort tag list

fzuellich:
Implemented enhancement #1527: "Tag list sort by name in View all tags".

 Added the `mixin wx.lib.mixins.listctrl.ColumnSorterMixin` that is implemented by `robotide.ui.tagdialogs.ViewAllTagsDialog`. This includes the creation of a new field `itemDataMap` and `sort_state` as well as a call to `self._tags_list.SetItemData` in `self._execute`.

The second commit ensures that the sort order is kept on refresh.

### DIFF
--- a/src/robotide/ui/tagdialogs.py
+++ b/src/robotide/ui/tagdialogs.py
@@ -32,12 +32,13 @@ class ViewAllTagsDialog(wx.Frame, listmix.ColumnSorterMixin):
         self.frame = frame
         self.tree = self.frame.tree
         self._controller = controller
-        self.itemDataMap = dict()
         self._results = utils.NormalizedDict()
         self.selected_tests = list()
         self.tagged_test_cases = list()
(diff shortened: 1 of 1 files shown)


yanne:
It seems that the sort is case sensitive. (try i.e. having tags `Tag`, `tag1` and `another tag`. I'd prefer the sort to be case insensitive, since the tags themselves are case insensitive from RF point of view.

fzuellich:
I wasn't aware of this fact. But I think the fix is made quick. I'll submit some changes later today.

yanne:
Thanks for the PR. I will merge this.

In the future, it would be cool if you can format your commit messages so that the first line is about 50 chars (summary of the change), followed by a new line and the longer explanation. This is a standard we try to follow, although not dogmatically.

Fixes #1527

PR #1570, opened 2016-01-21
Changes log tab name to 'RIDE Log' and other minor improvements.

HelioGuilherme66:
Better .format arguments.
Adds wx2.9 without announcing it.
Improves print and text with format().

### DIFF
--- a/src/robotide/__init__.py
+++ b/src/robotide/__init__.py
@@ -43,7 +43,7 @@ try:
     wxversion.select(supported_versions)
     import wx
 except ImportError as e:
-    if "no appropriate 64-bit architecture" in e.message.lower() and \
+    if "no appropriate 64-bit architecture" in "{0}".format(e).lower() and \
        sys.platform == 'darwin':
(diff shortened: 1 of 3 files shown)


yanne:
@HelioGuilherme66 have you actually tried to run RIDE with wxPython2.9? At some point, it had many incompatible APIs (even more that 3.0). I cannot really test it myself. From my point of view, it would be enough to just support 3.0 and skip 2.9 altogether.

pekkaklarck:
Definitely no need to support wxPython 2.9. It was a developement series.

HelioGuilherme66:
@yanne Yes I did see some errors in output. (I have some trouble with 2.8.12.1, there are no packages for Fedora, and local build fails with segmentation fault. So I am unable to test in 2.8.12.1) 
@pekkaklarck I did not know of that.
I will revert that change, and only test in 3.0.

yanne:
I merged this on the command line to be able to resolve some conflicts. I'll close this.

Issue #1579, opened 2016-01-26
Sort trailing numbers in tag names numerically in View All Tags


alexisbietti:
Related to enhancement #1527

Sort by tag name does a simple text comparison, which gives weird results when tag names have trailing numbers:

![2016-01-26 16_42_42-clipboard](https://cloud.githubusercontent.com/assets/220144/12585349/f9e3af62-c44b-11e5-8c79-10d64460c1cb.png)

This can be enhanced by parsing the tag name and considering any trailing number as a numeric value before sorting:

![view all tags](https://cloud.githubusercontent.com/assets/220144/12585054/9fffa6b4-c44a-11e5-9f52-ba952c0d08e1.png)

[SortTagByTrailingNumber.txt](https://github.com/robotframework/RIDE/files/105164/SortTagByTrailingNumber.txt)

HelioGuilherme66:
I did the PR #1580 using your code, since it is not available at your fork.
Tested with file at [Tags_test.robot](https://dl.dropboxusercontent.com/u/50838335/Robot%20Files/Tags_test.robot)

pekkaklarck:
I'm not sure is this a good idea. What about leading or intermediate numbers? A simple workaround for this problem is adding leading zeros like `Sprint-01`.

yanne:
I don't like this either, for the reasons @pekkaklarck mentioned. This opens up a door for supporting all the possible positions for numbers, which is a bit more complex case.

alexisbietti:
I don't mind if I'm the only one using this little modification. I agree the general case of supporting any position of number is really complex and not worth the effort. But trailing incrementing number are the norm in issue trackers such as Jira, and these tools don't insert leading zeroes to every past issue each time a new digit is reached just to fix sorting. Cheers.

alexisbietti:
By the way, thanks for taking the time to enter the PR @HelioGuilherme66 .

HelioGuilherme66:
Comparison of the Tag Sorting feature.
In the folowing images we compare the After and Before the code changes.
![after 1580_a](https://cloud.githubusercontent.com/assets/2866630/12833522/2754edea-cb99-11e5-9e16-4c403979dcd7.png)
after#1580_a.png - A - prefix numbering sorted numerically (ignores leading zeroes)
                   B - prefix numbering sorted alphanumerically (considers leading zeroes)
[...]

PR #1580, opened 2016-01-27
Improves Sort trailing numbers in tag names numerically ...

HelioGuilherme66:
in View All Tags #1579

Proposed by @alexisbietti.
(code is PEP8 compliant)


HelioGuilherme66:
Please hold merging this one. I will incorporate PR #1593 because they touch same file.

HelioGuilherme66:
Please merge. Thanks.

yanne:
This branch still contains 11 commits, most of which are noise. The commits should be squashed with `git rebase -i` possibly first rebasing on current master by doing `git fetch origin-ride && git rebase origin-ride/master` before this can be merged.

HelioGuilherme66:
I believe this is it! Merge me!

src/robotide/ui/tagdialogs.py
```diff
@@ -121,10 +130,10 @@ def _execute(self):
...
 
     def update_footer(self):
-        footer_string = "Total tests %d, Tests with tags %d, Unique tags %d," \
-                        " Currently selected tests %d" % \
-            (self.total_test_cases, len(self.tagged_test_cases),
-             self.unique_tags, len(self.selected_tests))
+        footer_string = "Total tests %d, Tests with tags %d, Unique tags %d, \
+Currently selected tests %d" % (self.total_test_cases,
+                                len(self.tagged_test_cases), self.unique_tags,
+                                len(self.selected_tests))
```
pekkaklarck: Is this just a formatting change? Should preferably be in a separate commit (or even pr). More importantly,  I don't think cutting a string like

```python
    "Example string \
here is."
```

is a good idea. I'd use either

```python
    "Example string " \
    "here is."
```

or probably

```python
    ("Example string "
     "here is.")
```

The latter is also what [PEP-8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) recommends:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

src/robotide/ui/tagdialogs.py
```diff
@@ -304,15 +312,14 @@ def add_tag(self, tag_to_tests):
...
-        insert statements.
+        the model does not. The entry in the model is referenced by a new id
+        generated by wx.NewId(). This prevents overwriting entries when
+        combining delete and insert statements.
 
         Args:
             index: An int marking the position to insert the row at.
             tag_to_tests: A tuple mapping tests(list, index 0) to a
-            tag(str, index 1).
+                          tag(str, index 1).
```
pekkaklarck: What happened here?

src/robotide/ui/tagdialogs.py
```diff
@@ -161,8 +170,7 @@ def _search_for_tags(self):
         isreversed = (self.sort_state[1] != 1)
         self.total_test_cases = len(self._test_cases)
         self._results = sorted(self._unique_tags.items(),
-                               key=lambda x: str(x[0]).lower,
-                               reverse=isreversed)
+                               key=lambda x: str(x[0]).lower, reverse=isreversed)
```
pekkaklarck: 1. I don't think this kind of changes should be done in general.

2. I actually like the original formatting better.

3. That `str(x[0])` call looks dangerous. It will fail if any `x[0]` contains non-ASCII characters. Are all these items always Unicode or byte strings? If yes, just using `x[0].lower` would work.

4. `item` would be a better name than `x` when sorting `.items()`.

src/robotide/ui/tagdialogs.py
HelioGuilherme66: It is just indentation for the doc.

src/robotide/ui/tagdialogs.py
```diff
@@ -100,6 +101,14 @@ def _make_bindings(self):
         self._tags_list.Bind(wx.EVT_LIST_ITEM_RIGHT_CLICK, self.OnRightClick)
         self._tags_list.Bind(wx.EVT_LIST_COL_CLICK, self.OnColClick)
 
+    def _tag_name_for_sort(self, tag_name):
+        # Make tag_name lowercase for the sorting algorithm only
+        tag_name = tag_name.lower()
+        # Compare trailing numbers as numeric values
+        m = re.search(r'([^0-9]*)([0-9]*)', tag_name)
```
pekkaklarck: According to the comment this regexp tries to to find trailing numbers, but the regexp itself actually finds *any* number. Is comment or code correct??

src/robotide/ui/tagdialogs.py
HelioGuilherme66: 1 and 2. It was an unnoticed change when squashing, I also like better as it was.

src/robotide/ui/tagdialogs.py
pekkaklarck: Oh, I see. It was indented to match `: ` on the above line. I guess it makes sense. Having changes like this in same commits as actual features is nevertheless confusing.

src/robotide/ui/tagdialogs.py
HelioGuilherme66: 3. Was indeed a bad piece of code. I have also updated the test file to have unicode chars.
4. Oops I forgot to change the var name (coming soon another forced commit!)

src/robotide/ui/tagdialogs.py
```diff
@@ -100,6 +101,14 @@ def _make_bindings(self):
         self._tags_list.Bind(wx.EVT_LIST_ITEM_RIGHT_CLICK, self.OnRightClick)
         self._tags_list.Bind(wx.EVT_LIST_COL_CLICK, self.OnColClick)
 
+    def _tag_name_for_sort(self, tag_name):
+        # Make tag_name lowercase for the sorting algorithm only
+        tag_name = tag_name.lower()
+        # Compare numbers as numeric values
+        m = re.search(r'([^0-9]*)([0-9]*)', tag_name)
+        g = m.groups()
+        return g[0], int(g[1] or 0)
```
pekkaklarck: This doesn't seem to handle tags like `1foo` and `1bar` or `foo1x` and `foo1y` correctly.

pekkaklarck:
Not really related to this PR, but looking at the code I noticed some weird code at https://github.com/robotframework/RIDE/blob/master/src/robotide/ui/tagdialogs.py#L144:

```python
        self._unique_tags = utils.NormalizedDict()
        self._tagit = utils.NormalizedDict()
        self._test_cases = list()
```

The weirdness is that none of these seem to be used outside the method where they are created and `self._tagit` is ever only populated. Do you guys see any reason why not to change the above to:

```python
        unique_tags = utils.NormalizedDict()
        test_cases = []
```

If that is a good change, could you @HelioGuilherme66 commit that along with the clean-ups you have spotted separately?

HelioGuilherme66:
Done the changes, except at self._tagit because it is used in Rename and Delete code.

yanne:
The `self._tagit` initialisation is definitely poor style, I'd prefer all instance variables to be declared in the `__init__` method. However, they are used and can't be just removed.

pekkaklarck:
Yeah, my bad. I wasn't looking for the whole file and didn't see other usages. Attributes should definitely be defined already in `__init__`, though.

`tagit` name is also pretty weird. It actually sounds like Finnish and should probably be renamed to `tags`.

src/robotide/ui/tagdialogs.py
HelioGuilherme66: The following images show the differences:
Before:
tag_sort_old_algorithm
After:
tag_sort_new_algorithm

Is it acceptable?

pekkaklarck: No, I don't think implementation that ignores everything after the first group of digits is acceptable.

HelioGuilherme66: Changed code to `m = re.search(r"(\w*)([0-9]*)", tag_name)` and now looks better (keeping the trailing numeric order):
tag_sort_mod_algorithm

HelioGuilherme66:
Final changes I hope ;)

src/robotide/ui/tagdialogs.py
```diff
@@ -100,6 +102,15 @@ def _make_bindings(self):
         self._tags_list.Bind(wx.EVT_LIST_ITEM_RIGHT_CLICK, self.OnRightClick)
         self._tags_list.Bind(wx.EVT_LIST_COL_CLICK, self.OnColClick)
 
+    def _tag_name_for_sort(self, tag_name):
+        # Make tag_name lowercase for the sorting algorithm only
+        tag_name = tag_name.lower()
+        # Compare numbers as numeric values
+        # m = re.search(r"([^0-9]*)([0-9]*)", tag_name)
+        m = re.search(r"(\w*)([0-9]*)", tag_name)
+        g = m.groups()
```
pekkaklarck: This code is still buggy. It ignores everything after the first group of digits like it did earlier. In other words, order of tags like `1a`, `1b` and `1c` is undefined. I'm not sure what's the best way to fix that.

pekkaklarck: This version is actually even more buggy than the old one. Now tags that contain special characters like `foo-1` are not handled correctly.

HelioGuilherme66:
Well, considering the initial objective, now the tags are listed in a better organized way (sorted). The test file is [Test_tags.robot](https://github.com/HelioGuilherme66/RIDE/raw/sort_tags_%231579/utest/resources/robotdata/view_all_tags/Tags_test.robot). Please say if you think that RF log file would benefit with this sorting variant (and should we put that under discussion in users group).

peterlck:
Hi,

I would like to unsubscribe from the thread and I have clicked on the
Unsubscribe link numerous time but to no avail.
Can someone help me to unsubscribe ?

HelioGuilherme66:
@peterlck You are getting this from GitHub not a User Group mailing list.
To stop receiving this, just click the Unwatch button in this project [RIDE](https://github.com/robotframework/RIDE/)

HelioGuilherme66:
@yanne @pekkaklarck @mkorpela 
Finally the code is ready to be merged.
Algorithm is simplified and unitest was added.
Includes fix #1621

src/robotide/ui/tagdialogs.py
```diff
@@ -100,6 +101,16 @@ def _make_bindings(self):
         self._tags_list.Bind(wx.EVT_LIST_ITEM_RIGHT_CLICK, self.OnRightClick)
         self._tags_list.Bind(wx.EVT_LIST_COL_CLICK, self.OnColClick)
 
+    def _tag_name_for_sort(self, tag_name):
+        mylist = [part if index % 2 == 0 else int(part) for index, part in
+                  enumerate(re.split(r'(\d+)', tag_name.lower()))]
```
pekkaklarck: This is a good way to split the name into comparable tokens. I would rename `mylist` to `tokens`, though.

src/robotide/ui/tagdialogs.py
```diff
@@ -100,6 +101,16 @@ def _make_bindings(self):
...
+        # Side effect from using regex, u'' will be removed from start or end
+        if mylist[0] == unicode(''):
+            del mylist[0]
+        if mylist[-1] == unicode(''):
+            del mylist[-1]
```
pekkaklarck: Empty strings shouldn't be removed. Keeping them enforces that the returned sort key always starts and ends with a string and is thus comparable. Sorting, for example, `('foo', 1)` and `(2, 'bar')` makes no sense (in Python 3 it even raises `TypeError`), but comparing `('foo', 1, '')` and `('', 2, 'bar')` is fine.

pekkaklarck:
1. The latest version looks pretty good. See my line note above for one place that ought to be fixed simply by removing few lines of code.

2. It is not a good idea to have unrelated changes in a commit or PR. There aren't too many of them so I'm fine leaving them this time.

3. Commits in this PR should be squashed.

HelioGuilherme66:
Squashed and final :)
Thanks!

PR #1584, opened 2016-01-31
Changed code to be PEP8 compliant and removed unnecessary method

fzuellich:
Changed almost all of the tree.py code to be PEP8 compliant (except the method `_create_node`, but I guess it is more readable like this).


yanne:
Thanks, good job.

pekkaklarck:
On a quick look changes looked very good. I would avoid empty lines inside methods and use them only between methods, classes, imports, etc. If a method grows so big that it requires separation to sections like that, it should generally be split to smaller pieces.

Issue #1416, opened 2015-06-03
When saving in the Text Edit screen, all test case checkboxes are cleared


robotframework-ride:
> *Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=1376) by billmat... on 28 Aug 2014*

RIDE unchecks any checked test cases when the file is saved in the Text Edit screen.  This makes it extra work when going between editing and re-running test cases.  It doesn't do this in the grid Edit screen.

I can see the challenge if the edits actually added or removed test cases, especially if it removed a checked test case.  However it would be nice if RIDE could remember which tests are checked, and at the point of saving/re-rendering the treeview, maintain the checks that are still relevant (i.e. the test case still exists by name).

This is with RIDE 1.3 on Windows 2008 <a href="https://code.google.com/p/robotframework-ride/source/detail?r=2">R2</a>.

yanne:
Fixed in #1586

PR #1586, opened 2016-02-02
Bugfix #1416: test case checkbox cleard upon save in textedit

fzuellich:
Fixing #1416 by removing call to `TreeSelectionController.unselect_all` after files change. As a consequence implementing a call to deselect nodes that are removed and thus maintaining a consistent model of selected tests. As far as I am aware of, no undo operation is provided for removed TestCases (at least none working on my local installation) and therefore did not need to be considered. A simple unit test is added to ensure new method in `TreeSelectionController` works as expected.

Also merged latest PEP8 related changes from master.

### DIFF
--- a/src/robotide/controller/ui/treecontroller.py
+++ b/src/robotide/controller/ui/treecontroller.py
@@ -176,6 +176,9 @@ class TestSelectionController(object):
     def is_empty(self):
         return not bool(self._tests)
 
+    def is_test_selected(self, test):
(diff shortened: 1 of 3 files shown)


yanne:
Looks good, but I'd prefer not having the merge commits and conflict resolution commits end up in history.

Could you do an interactive rebase on your branch, squashing all the commits into a single commit. That would make a cleaner history. The change itself is so small that it can be a single commit IMO.

Issue #1585, opened 2016-01-31
[DEV] Discussion about wx version detection and setup improvements


HelioGuilherme66:
This issue is to talk about improving the way wxPython is detected.

Currently the user does not have a way to select a older wx version (i.e. it always selects the highest version installed). The code enforcing this is in `__init__.py`:
```python
supported_versions = ["2.8"]

try:
    import wxversion
    from wxversion import VersionError
    if sys.platform == 'darwin':
        supported_versions.append("2.9")
    wxversion.select(supported_versions)
```
I propose that this mechanism is removed, to allow the user to select the wx version installed by:
- Setting `wx.pth` (in site-packages) with the default version name, or
- Passing an environmental variable, like `WXPYTHON` to override the default.

This suggestion also implies not using the module `wxversion`, which is the one forcing the highest installed version.

yanne:
I think this was originally implemented because there were some linux systems with wxPython 2.6 and 2.8 around. I think the wxVersion.select() call can be dropped, but I think we should still either give a warning or outright fail if we know an incompatible version if used.

HelioGuilherme66:
I'd like for you to try my branch [setup_improv](https://github.com/HelioGuilherme66/RIDE/tree/setup_improv) to help me getting a working ride_postinstall.py in Windows.
These are the changes in the branch:
- Removed distutils fallback
- Changed to setuptools and ez_install
- Replaced TK dialogs by wx dialogs
- Removed wxversion use
- Added a "--version" option to RIDE (I had the intention to report it in ride_postinstall.py, but did not due to major problem in Windows).
- Added the creation of a Desktop shortcut to RIDE in Linux (user must run `ride_postinstall -install` after setup or pip)
- Added the creation of a Desktop shell script to call RIDE in Mac OS X (user must run `ride_postinstall -install` after setup or pip)

The problems with Windows installer are:
- There is an error of loading the C++ DLL when ride_postinstall.py calls verify_install() which uses wx. The only way to workaround is not to use wx in ride_postinstall.py and Windows. I am almost certain that the distutils package to create the installer (`python setup.py bdist_wininst`) is getting a 32bit DLL that crashes on my x64 Windows 7.
- The ride_postinstall.py can only be called from the Windows installer (due to special function call). This prevents me to have a similar solution as for Linux and OS X.
[...]

PR #1589, opened 2016-02-08
Modernized setup, relax wx verification, new --version option, post install Desktop shortcut

HelioGuilherme66:
Documentation will be updated later.

List of significant changes:
Creates desktop shortcuts for all platforms.
Adds a --version option to RIDE.
Removed wxversion checks from \_\_init\_\_.py. Added warning about proper wx version to use.
Completed ride_postinstall for Linux.
Changed setup.py to use setuptools and ez_setup.
Corrected windows detection, and arguments for post_install.
Create a Desktop shortcut to Linux script starting RIDE.
This is a minimal solution to have shortcut in Mac OS X.
Calls ride_postinstall.py after install.
Added ride_postinstall to run after setup.py or pip.
Removed verify_install when in Windows. Now ride_postinstall can be run standalone in all operative systems.
Asks to create Desktop shortcut in Windows.
Executable installer in Windows shows error but completes.
Introduces a new dependency pywin32 (packages only from sourceforge).
Fixed user detection on installs with "sudo".

ride_postinstall.py is not compatible with Python 2.6.

### DIFF
--- a/ride_postinstall.py
+++ b/ride_postinstall.py
@@ -1,31 +1,158 @@
+#!/usr/bin/env python
+# encoding=utf-8
+
 import sys
 from os.path import exists, join
-from Tkinter import Tk
-from tkMessageBox import askyesno
+
+__doc__ = """
+Usage: ride_postinstall.py <-install|-remove>
+""".strip()
+# TODO: Add -remove, to remove desktop shortcut
 
 
 def verify_install():
     try:
-        import wxversion
-    except ImportError:
-        print "No wxPython installation detected!"
-        print ""
-        print "Please ensure that you have wxPython installed before running RIDE."
-        print "You can obtain wxPython 2.8.12.1 from http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/"
+        from wx import version
+    except ImportError as err:
+        print("No wxPython installation detected!")
+        print("")
+        print("Please ensure that you have wxPython installed before running \
+RIDE.")
+        print("You can obtain wxPython 2.8.12.1 from \
+http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/")
+        sys.exit(1)
     else:
-        print "Installation successful."
+        print("Installation successful.")
 
 
-def create_desktop_shortcut():
-    Tk().withdraw()
-    link = join(get_special_folder_path("CSIDL_DESKTOPDIRECTORY"), 'RIDE.lnk')
-    icon = join(sys.prefix, 'Lib', 'site-packages', 'robotide', 'widgets', 'robot.ico')
-    if exists(link) or askyesno('Setup', 'Create desktop shortcut?'):
-        create_shortcut('pythonw', "Robot Framework testdata editor", link,
-                        '-c "from robotide import main; main()"', '', icon)
-        file_created(link)
+def _askyesno(title, message):
+    import wx
+    _ = wx.App()
+    parent = wx.Frame(None, size=(0, 0))
+    parent.CenterOnScreen()
+    dlg = wx.MessageDialog(parent, message, title, wx.YES_NO |
+                           wx.ICON_QUESTION)
+    result = dlg.ShowModal() == wx.ID_YES
+    dlg.Destroy()
+    return result
 
 
-if sys.argv[1] == '-install':
-    verify_install()
-    create_desktop_shortcut()
+def _askdirectory(title, initialdir):
+    import wx
+    _ = wx.App()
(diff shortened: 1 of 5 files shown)


pekkaklarck:
1. What was wrong with the old shortcut creation code on Windows? Must be something serious to warrant adding dependency to pywin32.

2. On what Linux distributions have you tested the code?

3. What are the benefits of including ez_setup compared to just having setuptools as a normal dependency? Most users have it already and others can easily pip install it.

4. Why not to have shortcuts creation code in a submodule that can be executed like `python -m robotide.postinstall`? Installing separate `ride_postinstall.py` into `/usr/bin` or elsewhere doesn't sound good outside Windows.

HelioGuilherme66:
1. The old shortcut creation code on Windows, only runs with installer and not standalone and by pip. It contains a deprecated call to DLL. The changes I did still allows to create and use installer, even if we will focus on pip.

2. I have tested on Fedora 22 x64 KDE/Plasma and CentOS 7 x64 Gnome Classic. I may test on OpenSUSE 13.1 x64 Gnome Unity and CentOS 6.7 i386 Gnome Classic (looks like I need to have a Debian based test machine ;) ).

3. I included ez_setup because I thought it was the "standard" with setuptools, for a first time installation. I am OK in removing it.

4. I started by modifying existing code and configuration, I am OK in changing it to a submodule.
> It may be of value to design desktop shortcut creation like in the following:
>         a. if in user's RIDE properties is "first run for this version" and no desktop shortcut exits then:
>                 Ask user if wants to create desktop shortcut (showing checkbox inactive to "Don't ask me on next start" ).
>         b. Have a menu entry to the create desktop shortcut function. (nice to have: configuring adding argument to a directory or .robot

HelioGuilherme66:
@yanne What do you think about this PR?

At 3. should we remove ez_setup? 
(I don't know if it is related with this, but now pip install forces to upgrade setuptools to latest. This means that usually people have version 0.6 and must upgrade to version 20.)
Should we adopt the strategy proposed at 4.?

yanne:
I am so out of touch concerning Python packaging and installation that I don't have any strong opinions about this. I would be very grateful if @pekkaklarck has the time to review this, since I simply do not have the competence.

HelioGuilherme66:
Thanks for the answer.
I will now proceed with an alternative proposal, based my own and @pekkaklarck 's comments.

HelioGuilherme66:
Continuing discussing progress from [comment](https://github.com/robotframework/RIDE/issues/1585#issuecomment-183037949).

pekkaklarck:
Few general comments:

1. This PR contains too many unrelated changes. At least changes related to wxversion should be in a separate PR.

2. Adding pywin32 as a dependency to get shortcuts is, in my opinion, not acceptable. I still don't understand what was wrong with the old approach other than it only was executed when using the binary installer. If that doesn't really work, it should be studied could ctypes be used to call the needed Windows APIs instead.

3. The Linux solution doesn't seem to be generic. Considering how many Linux distributions there are, I'm not sure is a generic solution possible or worth the effort. If something is implemented, it definitely should support both RedHat and Debian based distros. Also, it is `Työpöytä` in Finnish.

4. I'm in general not sure is it a good idea to try using pip to create shortcuts and other platform dependent configuration. This at least until pip supports such functionality out-of-the-box. If better platform integration is needed, I think it would be better to use something like py2exe on Windows, create RPM and DEB for Linux, etc.

HelioGuilherme66:
1. Removing wxversion was an initial objective of this PR.
2. The goal is to stop using binary installer.
3. Only missing to test in Debian based. Will correct the word, was a typo.
4. Is being developed in the branch [desktopshort](https://github.com/HelioGuilherme66/RIDE/tree/desktopshort) as I mentioned. User will run desktop creator after RIDE is running.

pekkaklarck:
If the initial objective was removing wxversion, this PR shouldn't have ever got any more content. If that is a good idea (I don't know), it could be merged separately and other topics discussed later. This applies also to your desktopshorts branch which seems to contain lot of other functionality than just shortcuts.

HelioGuilherme66:
I would need a decision on: 
a) we will continue to produce the windows installer (if so we need a wrapper for postinstall module);
b1) installation of shortcut automatically after pip, setup, installer, OR
b2) installation of shortcut automatically after starting RIDE,
c) installation of shortcut manually by calling module or from RIDE (this is already in place)

HelioGuilherme66:
Closing in favor of #1597.
Separate PRs for RIDE version and wxversion remove.

PR #1595, opened 2016-02-14
Adds --version option to RIDE.

HelioGuilherme66:
Uses print function.
Replaces wx.PySimpleApp by wx.App to avoid future deprecation.
(Code is PEP8 compliant)

### DIFF
--- a/src/robotide/__init__.py
+++ b/src/robotide/__init__.py
@@ -14,7 +14,7 @@
 
 """RIDE -- Robot Framework test data editor
 
-Usage: ride.py [--noupdatecheck] [--debugconsole] [inpath]
+Usage: ride.py [--noupdatecheck] [--debugconsole] [--version] [inpath]
 
 RIDE can be started either without any arguments or by giving a path to a test
 data file or directory to be opened.
@@ -22,6 +22,8 @@ data file or directory to be opened.
 To disable update checker use --noupdatecheck.
 
 To start debug console for RIDE problem debugging use --debugconsole option.
+
+To see RIDE's version use --version.
 
 RIDE's API is still evolving while the project is moving towards the 1.0
 release. The most stable, and best documented, module is `robotide.pluginapi`.
@@ -69,6 +71,14 @@ def main(*args):
     if len(args) > 3 or '--help' in args:
         print __doc__
         sys.exit()
+    if '--version' in args:
+        try:
+            from robotide import version
+        except ImportError:
+            print("Error getting RIDE version!")
+            sys.exit(1)
+        print(version.VERSION)
+        sys.exit(0)
     try:
         _run(inpath, not noupdatecheck, debug_console)
     except Exception:


src/robotide/__init__.py
```diff
@@ -33,7 +35,8 @@
 
 errorMessageTemplate = Template("""$reason
 You need to install wxPython 2.8.12.1 with unicode support to run RIDE.
-wxPython 2.8.12.1 can be downloaded from http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/""")
+wxPython 2.8.12.1 can be downloaded from http://sourceforge.net/projects/wxpyt\
+hon/files/wxPython/2.8.12.1/""")
```
pekkaklarck: Cutting URLs like this doesn't make sense. We don't need to follow PEP-8 that pedantically. Cutting the line before the URL would be OK, though.

src/robotide/__init__.py
```diff
@@ -82,7 +93,7 @@ def _parse_args(args):
         return False, False, None
     noupdatecheck = '--noupdatecheck' in args
     debug_console = '--debugconsole' in args
-    inpath = args[-1] if args[-1] not in ['--noupdatecheck', '--debugconsole'] \
+    inpath = args[-1] if args[-1] not in ['--noupdatecheck', '--debugconsole']\
```
pekkaklarck: Why this change? At least I prefer whitespace before `\` denoting line continuation.

HelioGuilherme66:
Done changes. Ready to merge?

src/robotide/__init__.py
```diff
@@ -109,15 +120,17 @@ def _show_old_wxpython_warning_if_needed(parent=None):
...
     message = ('RIDE officially supports wxPython 2.8.12.1. '
                'Your current version is %s.\n\n'
-               'Older wxPython versions are known to miss some features used by RIDE. '
+               'Older wxPython versions are known to miss some features used b\
+y RIDE. '
                'Notice also that wxPython 3.0 is not yet supported.\n\n'
                'wxPython 2.8.12.1 packages can be found from\n'
-               'http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/.'
+               'http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12\
+.1/.'
```
pekkaklarck: Cutting lines like this makes absolute no sense. The source code gets a lot harder to read. Using a multiline string would be a better solution here.

src/robotide/__init__.py
```diff
@@ -67,8 +70,16 @@
 def main(*args):
     noupdatecheck, debug_console, inpath = _parse_args(args)
     if len(args) > 3 or '--help' in args:
-        print __doc__
+        print(__doc__)
         sys.exit()
+    if '--version' in args:
+        try:
+            import version
```
pekkaklarck: This should be `import .version` or `from robotide import version`. Relative imports are always confusing and, in case it matters, they aren't Python 3 compatible.

pekkaklarck:
General comments:

- PR *should not* contain both a new feature (e.g. `--version`) and code cleanup to unrelated lines.
- Separate cleanup PRs are OK, although trivial whitespace changes just add noise to the project history.
- Cutting works from the middle just to keep lines below 80 characters makes *absolute no sense* and only makes the code a lot harder to read. Even PEP-8 says that "A Foolish Consistency is the Hobgoblin of Little Minds" and "Readability counts".
- If `print('foo')` is used in any non-trivial code, there should also be `from __future__ import print_function`. Although `print('foo')` works also without it, `print('foo', 'bar')` and print('foo', file=sys.stderr)` don't.

I let @yanne to merging when he thinks this code is ready.

yanne:
I don't actually mind having cleanup and new features in the same PR. They might still be separate commits to distinguish between changes needed for the feature and the cleanup. 

Other than that, I agree with @pekkaklarck's comments, and this PR needs a little cleanup.

As a technical git suggestion, you can use `commit --amend` and `push --force` to change the contents of this PR, because no other work is based on this.

pekkaklarck:
@yanne is correct that it's more important that individual commits don't contain unrelated changes. I still thing clean-ups are is typically better handled as separate PRs. That makes merging them easy and they are not left waiting if the other changes in the PR require still more work.

HelioGuilherme66:
Just changed the relative import.
The rest of formatting fixes goes in remove wxversion PR.
Thanks.

HelioGuilherme66:
Sorry for pressing you guys. I really want to move on with fixes and features. This PR now only have the --version add. Then in other PR the remove_wxversion, will change to print() and PEP8. All these changes are needed for the installer PR and to have wxPython 3 (and 2.8.12) working, which I want to have in master.
Let's merge and close this one? Thanks.

yanne:
This is now okay.

PR #1596, opened 2016-02-14
Removes wxversion

### LINKED ISSUE #1569: wxversion 2.8 does not use wxversion to check wx version in centos
wxversion 2.8 does not use wxversion to check wx version in Centos Server.

And another API for wx.version() to get the current wx version.
So, the wx version detect method in robotide.**init** may need change.

HelioGuilherme66:
This simplifies wx use, because user may choose what wx version to use. Proper warnings are shown if version is not supported.
Fixes #1569.

### DIFF
--- a/src/robotide/__init__.py
+++ b/src/robotide/__init__.py
@@ -109,15 +109,15 @@ def _show_old_wxpython_warning_if_needed(parent=None):
     if wx.VERSION >= (2, 8, 12, 1, ''):
         if wx.VERSION > (2, 8, 12, 1, ''):
             title = "Please be aware of untested wxPython installation"
-            message = ("RIDE officially supports wxPython 2.8.12.1."
-                       "Your current version is %s."
+            message = ("RIDE officially supports wxPython 2.8.12.1. Your "
+                       "current version is %s."
                        "\n"
                        "There are significant changes in newer wxPython "
                        "versions. Notice that RIDE is still under development "
                        "for wxPython 3.0.2 and newer (wxPython-Phoenix). "
-                       "wxPython 2.8.12.1 packages can be found from "
-                       "http://sourceforge.net/projects/wxpython/files/"
-                       "wxPython/2.8.12.1/."
+                       "wxPython 2.8.12.1 packages can be found from http://"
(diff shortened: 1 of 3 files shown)


HelioGuilherme66:
I did try to use """...""" but could not get rid of "\n" for empty lines. So this will be the last changes I do on this PR. If you still don't agree with this multiline style, I need to know how you want, because there will be changes in branches compatible also with wxPython 3.0.2 and wxPhoenix (>=3.0.3) which I am already working on.

yanne:
I just tested this now, and this implementation does not gracefully handle wxPython 2.8.12.1 and 3.0.x being installed at the same time. The problem is the `wx.CallAfter` used to display the error message. I think it would be best to revert that change, and simply refuse to start RIDE with incompatible wxPython version.

HelioGuilherme66:
I also did find those errors at the exit of RIDE with wx 2.8.12.1, but I do not believe that the fault is in `wx.CallAfter`, at least in that file. I tested with a merge over master with this branch and upstream/wxPython3-compatibility (and a fix for wx.Colour in 3.0.2) and I do not get errors. I will try to find what code is causing those messages.

yanne:
What I meant is that because the error is displayed using `wx.CallAfter`, too much if the initialisation gets done with incompatible wxPython version. If you remove the wx.CallAfter, and just exit on that line, there would be no such problem.

HelioGuilherme66:
I found the commit that was causing the error messages and deprecation warnings. They were at the Publisher methods. Please merge

yanne:
That's not my point.

We are talking about wx 2.8.12.1 compatible version here. If I try to run it with wxPython 3.0, it should give me a clear error message and exit. It does not happen with this version. We don't want to take any wxPython3 compatibility fixes in the current master.

yanne:
With wxPython 2.8, this branch is fine. If the wxPython3 issue can be resolved, I think this is ready for merging.

yanne:
I cleaned up the commits and pushed them into master myself. This can thus be closed.

Be sure to make a rebase on current master on any branches that may include commits from this branch. They may have merge conflicts.

PR #1597, opened 2016-02-14
Creates desktop shortcuts for all platforms.

### LINKED ISSUE #1582: Use `setuptools', not `distutils` for installation
`distutils` is deprecated and `setuptools` should be used instead. If I've got it right, installations using `pip` will always injecting `setuptools` into the process somehow, but when installing with `python setup.py install` that's not the case. `pip` is also trying to drop the special `distutils` support, although it seems they needed to revert that now: https://github.com/pypa/pip/issues/3384

Robot Framework has implemented this with the following code after normal `distutils` imports in `setup.py`. Apparently `setuptools` automagically monkeypatches `distutils` somehow.

``` python
try:
    import setuptools    # use setuptools when available
except ImportError:
    pass
```

Alternatively it should be possible to just import needed entry points directly from `setuptools`. It is also questionable is a fallback to `distutils` needed. At least `request` has recently dropped it: https://github.com/kennethreitz/requests/pull/2973

HelioGuilherme66:
Removes distutils use. Fixes  #1582.
Corrected windows detection, and arguments for post_install.
Create a Desktop shortcut to Linux script starting RIDE.
This is a minimal solution to have shortcut in Mac OS X.
Calls ride_postinstall.py after install.
Added ride_postinstall to run after setup.py or pip.
Now ride_postinstall can be run standalone in all operative systems.
Asks to create Desktop shortcut in Windows.
Introduces a new dependency pywin32 (packages only from sourceforge).
Fixed user detection on installs with "sudo".

ride_postinstall.py is not compatible with Python 2.6.

Fix for linux desktop directory not localized.
Corrected shebang in ride.py.

Transformed ride_postinstall.py into a module. Tested in Linux: Ubuntu,
Fedora, CentOS, OpenSUSE.
To run: python -m robotide.postinstall -install

Now Desktop Shortcut creation can be started from command line and GUI.

Adds a wrapper for postinstall module for Windows installer.
Use: ride_postinstall.py -install

Add options -q and -f.
Code clean up.
(code is PEP8 compliant)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
-# setuptools is required by mock
-setuptools
-nose
-mock
-invoke
-Pygments
+wxPython
+Pygments # This enables syntax highlighted in Text Editor
+robotframeworklexer # Updates lexer to newer RF 3.1.1
 robotframework
+Pypubsub==3.3.0 ; python_version <= '2.7'
+Pypubsub ; python_version > '2.7'
+# Only for Windows to have a shortcut created (install before RIDE)
+Pywin32; sys_platform == 'win32'
--- a/setup.py
+++ b/setup.py
@@ -1,29 +1,36 @@
-import os
-from os.path import abspath, join, dirname, isdir, isfile
-from distutils.core import setup
+#!/usr/bin/env python
+#  Copyright 2008-2015 Nokia Networks
+#  Copyright 2016-     Robot Framework Foundation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


pekkaklarck:
This PR contains too many unrelated changes. Stuff like adding `--version`, removing `wxversion`, PEP-8 compatibility, etc. should all be in separate PRs. I have mentioned and explained this problem so many times that I'm not going to spent time explaining it again. I don't think this PR can be merged or even properly reviewed in the current form.

HelioGuilherme66:
I have created separate PRs for version and wxversion removal.

Issue #1598, opened 2016-02-16
RIDE fails to load (traceback generated) if a plugin fails


memoselyk:
With RIDE 1.5.

I came across this issue, while trying to open two RIDE instances.
The `LogPlugin` failed to load, due to a shared file being _used by other process_, and it caused RIDE to fail, with the following traceback

> Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\robotide\__init__.py", line 73, in main
    _run(inpath, not noupdatecheck, debug_console)
  File "C:\Python27\lib\site-packages\robotide\__init__.py", line 99, in _run
    ride = RIDE(inpath, updatecheck)
  File "C:\Python27\lib\site-packages\robotide\application\application.py", line 42, in __init__
    wx.App.__init__(self, redirect=False)
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 7981, in __init__
    self._BootstrapApp()
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 7555, in _BootstrapApp
    return _core_.PyApp__BootstrapApp(*args, **kwargs)
  File "C:\Python27\lib\site-packages\robotide\application\application.py", line 55, in OnInit
    coreplugins.get_core_plugins())
  File "C:\Python27\lib\site-packages\robotide\application\pluginloader.py", line 30, in __init__
    standard_classes + self._find_classes(load_dirs) ]
  File "C:\Python27\lib\site-packages\robotide\application\pluginconnector.py", line 23, in PluginFactory
    return BrokenPlugin(utils.get_error_details(), plugin_class)
  File "C:\Python27\lib\site-packages\robotide\application\pluginconnector.py", line 69, in __init__
    + 'Error: ' + error
TypeError: cannot concatenate 'str' and 'tuple' objects

The problem is caused because `robot`'s `get_error_details` method returns a tuple with exception message and the produced traceback, since RF 2.9.

memoselyk:
The triggering failure of this problem, makes me wonder if that should be more gracefully handled.

> WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users
28d5ddd-48e3-42db-b97f-e1172c48d52f-ride.log'
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\robotide\application\pluginconnector.py", line 21, in PluginFactory
    plugin = plugin_class(application)
[...]

PR #1599, opened 2016-02-16
Update BrokenPlugin to RF 2.9's get_error_details method

memoselyk:
In RF 2.9, Robot Framework's get_error_details method was updated to
return a tuple with message and traceback.

This fixes issue #1598.


yanne:
Thanks for the patch

PR #1600, opened 2016-02-16
Fixes #1556, by ignoring starting Gherkin keywords.

### LINKED ISSUE #1556: Rename GIVEN WHEN THEN keywords does not work properly
Usage scenario:

1. I create a new test in the _given_ / _when_ / _then_ format. For example
 * Given a valid username and password
 * When I try to login with those credential
 * Then I successfully login
2. I right-click on each step (G/W/T) and select _**Create Keyword**_. Each Keyword is _correctly_ created without the G/W/T words, that is
 * a valid username and password
 * I try to login with those credential
 * I successfully login
3. Suppose I want to rephrase the _given_ , so I right-click on the keyword _at the usage point_ and select _**Rename Keyword**_ .

**Now, while the usage of the keyword is renamed, the actual keyword is not**.

Renaming works if I select _**Rename**_ on the Keyword object (in the left tree-panel).

I am using RIDE 1.5 with Python 2.7.10.

HelioGuilherme66:
There is a known bug. This is the comment in code:
-# When we change non-gherkin to gherkin, the keyword changes too.
-# The workaround is not to Rename keyword, but only edit field.

### DIFF
--- a/src/robotide/controller/commands.py
+++ b/src/robotide/controller/commands.py
@@ -290,8 +290,8 @@ class RenameKeywordOccurrences(_ReversibleCommand):
         self._occurrences = None
 
     def _check_gherkin(self, new_name, original_name):
-        was_gherkin, keyword_name = self._get_gerkin(original_name)
-        is_gherkin, new_keyword_name = self._get_gerkin(new_name)
+        was_gherkin, keyword_name = self._get_gherkin(original_name)
+        is_gherkin, new_keyword_name = self._get_gherkin(new_name)
         if was_gherkin and not is_gherkin:
             keyword_name = original_name
         if not was_gherkin and is_gherkin:
@@ -306,7 +306,7 @@ class RenameKeywordOccurrences(_ReversibleCommand):
                 keyword_name = original_name
         return keyword_name, new_keyword_name
 
-    def _get_gerkin(self, original_name):
+    def _get_gherkin(self, original_name):
         keyword_value = re.sub(self._gherkin_prefix, '', original_name)
         value_is_gherkin = (keyword_value != original_name)
         return value_is_gherkin, keyword_value
--- a/src/robotide/controller/commands.py
+++ b/src/robotide/controller/commands.py
@@ -15,6 +15,7 @@
 from itertools import chain
 import time
 import os
+import re
 
 from robotide.namespace.embeddedargs import EmbeddedArgsHandler
 from robotide.publish.messages import RideSelectResource, RideFileNameChanged,\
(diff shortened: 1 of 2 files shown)


src/robotide/controller/commands.py
```diff
@@ -279,12 +280,35 @@ class NullObserver(object):
 class RenameKeywordOccurrences(_ReversibleCommand):
 
     def __init__(self, original_name, new_name, observer, keyword_info=None):
-        self._original_name = original_name
-        self._new_name = new_name
+        keyword_name, new_keyword_name = self._check_gherkin(new_name,
```
yanne: These local variables are redundant. You can do `self._original_name, self._new_name = self._check_gherkin(...)`

src/robotide/controller/commands.py
```diff
@@ -279,12 +280,35 @@ class NullObserver(object):
...
+                                                             original_name)
+        self._original_name = keyword_name
+        self._new_name = new_keyword_name
         self._observer = observer
         self._keyword_info = keyword_info
         self._occurrences = None
 
+    def _check_gherkin(self, new_name, original_name):
+        _gherkin_prefix = re.compile('^(Given|When|Then|And|But) ',
```
yanne: local variable names do not need underscore prefixes

yanne: On a related note, I think the regex should be a class variable, since then it would only be compiled once.

src/robotide/controller/commands.py
```diff
@@ -279,12 +280,35 @@ class NullObserver(object):
...
+                                     re.IGNORECASE)
+        keyword_name = re.sub(_gherkin_prefix, '', original_name)
+        was_gherkin = (keyword_name != original_name)
```
yanne: I think these two lines should be extracted as a helper method, since they are repeated below.

src/robotide/controller/commands.py
```diff
@@ -279,12 +280,35 @@ class NullObserver(object):
...
+        is_gherkin = (new_keyword_name != new_name)
+        if was_gherkin and not is_gherkin:
+            keyword_name = original_name
+        if not was_gherkin and is_gherkin:
+            # When we change non-gherkin to gherkin, the keyword changes too.
+            # The workaround is not to Rename keyword, but only edit field.
+            new_keyword_name = new_name
+        if was_gherkin and is_gherkin:
+            # Check if the first word has changed
+            if original_name.split(' ', 1)[0].lower() != new_name.split(
```
yanne: You could use _gherkin_prefix.match.groups() to extract the gherkin prefix.

yanne:
I made some line notes. Also, would there be a way to write unit tests for this, since this is a bit a complex logic?

HelioGuilherme66:
Please see if you agree with changes. Thanks.

yanne:
I think this looks good.

Issue #1605, opened 2016-02-22
Find Usages not working for variables definitions


HelioGuilherme66:
If we right click in a cell having a variable definition or assignment there should show a choice dialog.
(the dialog is not shown and goes strait to results, if there is no = sign)
This is the dialog after the fix (coming in a PR soon):
![find_usage_variable_definition_ok](https://cloud.githubusercontent.com/assets/2866630/13206530/7d2377ee-d8f7-11e5-8c2a-64aeb771775e.png)

HelioGuilherme66:
I am going to propose a PR with the following solutions:
- Removed repeated variable names from cell content but keeping occurrence order
![find_usage_variable_definition_ok1_dialog](https://cloud.githubusercontent.com/assets/2866630/13289434/f119f18e-db07-11e5-9ce7-dca9b40868a6.png)

- There is a bug in the dialog which prevents to show the '&' in dictionaries (but results are OK). I opted to show \x26{varname} instead of just {varname}.
![find_usage_variable_definition_ok_results](https://cloud.githubusercontent.com/assets/2866630/13289450/013533e4-db08-11e5-8a2e-ad64eaac38b2.png)

yanne:
wxPython Controls (`wx.Control` subclasses, which `wx.RadioBox` is) have a feature of using ampersands to define mnemonics. You can escape the ampersand with a second (i.e. `&&` will render one ampersand). You could also use http://wxpython.org/Phoenix/docs/html/Control.html#Control.EscapeMnemonics method to escape all labels. I think the latter would be best, since then you don't have to manually search the ampersands.

HelioGuilherme66:
I tried all those options. It just happens that `wx.RadioBox` does not allow to escape the ampersand. I was not able to override `wx.RadioBox`, to use own SetString for the `choices` parameter. Please help me figuring out this,

yanne:
Fixed by #1606

PR #1606, opened 2016-02-24
Fix finding usages of variables (#1605).

HelioGuilherme66:
Not possible to show '&' in dictionary variables wx.RadioBox. (but possible with wx.RadioButton, see https://github.com/wxWidgets/Phoenix/issues/39 )
(includes PEP8 changes)

### DIFF
--- a/src/robotide/editor/kweditor.py
+++ b/src/robotide/editor/kweditor.py
@@ -16,10 +16,11 @@ import wx
 from wx import grid
 
 from robotide.context import IS_MAC
-from robotide.controller.commands import ChangeCellValue, ClearArea, PasteArea,\
-    DeleteRows, AddRows, CommentRows, InsertCells, DeleteCells, UncommentRows,\
-    Undo, Redo, RenameKeywordOccurrences, ExtractKeyword, AddKeywordFromCells,\
-    MoveRowsUp, MoveRowsDown, ExtractScalar, ExtractList, InsertArea
+from robotide.controller.commands import ChangeCellValue, ClearArea,\
+    PasteArea, DeleteRows, AddRows, CommentRows, InsertCells, DeleteCells,\
+    UncommentRows, Undo, Redo, RenameKeywordOccurrences, ExtractKeyword,\
+    AddKeywordFromCells, MoveRowsUp, MoveRowsDown, ExtractScalar, ExtractList,\
+    InsertArea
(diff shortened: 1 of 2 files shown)


src/robotide/editor/kweditor.py
```diff
@@ -728,13 +730,17 @@ def Clone(self):
...
                            style=wx.DEFAULT_DIALOG_STYLE)
         self.caption = "Please select what you want to check for usage"
-        variables = utils.find_variable_basenames(cellvalue)
+        variables = variablematcher.find_variable_basenames(cellvalue)
+        variables = variablematcher.find_unique(variables)
         self.choices = [(False, cellvalue)] + [(True, v) for v in variables]
+        # Bug in wx.RadioBox never shows the '&'
         self.choices_string = ["Complete cell content"] + \
-                              ["Variable " + var for var in variables]
+                              ["Variable " + var.replace("&", "\\x26") for var
```
yanne: Change the replace string here from  "\\x26" to "&&" and it should work

src/robotide/editor/kweditor.py
```diff
@@ -728,13 +730,17 @@ def Clone(self):
 class ChooseUsageSearchStringDialog(wx.Dialog):
 
     def __init__(self, cellvalue):
+        from collections import OrderedDict
         wx.Dialog.__init__(self, None, wx.ID_ANY, "Find Where Used",
                            style=wx.DEFAULT_DIALOG_STYLE)
         self.caption = "Please select what you want to check for usage"
-        variables = utils.find_variable_basenames(cellvalue)
+        variables = variablematcher.find_variable_basenames(cellvalue)
+        variables = variablematcher.find_unique(variables)
```
yanne: I'd prefer to chain these function calls instead of reassigning a variable with the same name.

src/robotide/editor/kweditor.py
```diff
@@ -728,13 +730,17 @@ def Clone(self):
 class ChooseUsageSearchStringDialog(wx.Dialog):
 
     def __init__(self, cellvalue):
+        from collections import OrderedDict
```
yanne: This line is unneeded and should be removed

HelioGuilherme66:
Code is ready to be merged. The "&&" bug was confirmed for GTK case (Linux) it will work as expected when wxWidgets contains the fix.

yanne:
Fixed by #1606

PR #1613, opened 2016-03-06
Fixes viewalltags dialog to show tags with unicode characters.

HelioGuilherme66:
The View All Tags dialog would not show if there are Tags with unicode characters. This PR fixes this.


yanne:
Thanks, this works perfectly.

PR #1621, opened 2016-03-17
Fixes delete and rename tags from View All Tags dialog (regression)

HelioGuilherme66:
Fixes delete and rename tags from View All Tags dialog due to renaming of variable self._tagit to self._tags.

This was introduced with commit cda9fedaf2f0ecc89b30af390f60725cef7dcf08


HelioGuilherme66:
Closing because this fix is included in #1580.

PR #1610, opened 2016-02-29
Wx python3 compatibility

HelioGuilherme66:
Merge from master.

PR #1632, opened 2016-05-03
1629 - RIDE will not run in ubuntu 16.04

timjohncurtis:
- Ubuntu 16.04 runs the 3.x wx libraries from the xenial repositories,
  and it seems the wx.RGBcolour object has been deprecated. So I have
  converted it to use the base wx.Colour object, converting the hex
  to 3 part RGB
- 'invoke test' now passes all tests with this change in place

### DIFF
--- a/src/robotide/ui/tree.py
+++ b/src/robotide/ui/tree.py
@@ -304,7 +304,7 @@ class Tree(treemixin.DragAndDrop, customtreectrl.CustomTreeCtrl,
 
         if isinstance(controller, ResourceFileController):
             if not controller.is_used():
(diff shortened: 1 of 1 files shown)


sdesbure:
+1 with this pull request so I can use RIDE on debian stretch!

HelioGuilherme66:
We have this fixed in PR #1610. You may install from https://github.com/HelioGuilherme66/RIDE/releases/tag/v2.0a2 with command:
`pip install -U https://github.com/HelioGuilherme66/RIDE/archive/v2.0a2.tar.gz`

PR #1663, opened 2016-10-24
Added context menu items

StefanIGit:
- for remove read only attribute (I think its cross plattfrom, but only
  tested on win 7 x64)
  -> could be improved the be active only on read only files
- for open containing folder ( only working on win32, , but only tested
  on win 7 x64)

### DIFF
--- a/src/robotide/controller/commands.py
+++ b/src/robotide/controller/commands.py
@@ -485,7 +485,17 @@ class DeleteFile(_Command):
         context.remove_from_filesystem()
         context.remove()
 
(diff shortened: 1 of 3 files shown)


HelioGuilherme66:
@StefanIGit I confirm that changing the read only attribute is also working on Linux. If you would like to try my changes to your PR, please see https://github.com/HelioGuilherme66/RIDE/tree/context_menu_items_%231663 . You'll have to pick only the filecontrollers.py changes.

It would be interesting if anyone could confirm if calling "Finder" is working on Mac OS.

PR #1679, opened 2017-01-20
Adds "context" to invoke>=0.13 methods.

HelioGuilherme66:
This is required for newest invoke.


HelioGuilherme66:
@pekkaklarck @yanne This is the same fix as done on `robotframework`.

PR #1698, opened 2017-05-29
Direct pythonpath order

gaal-dev:
Hello.

There is an error in src/robotide/namespace/namespace.py. For example pythonpath contains [u'[path1]', u'[path2]', u'[path3]']. The running script gets this pythonpath but when a library is imported namespace.py: _set_pythonpath reverses this order.

`    def _set_pythonpath(self):
        """Add user configured paths to PYTHONAPATH.
        """
        for path in self._settings.get('pythonpath', []):
            if path not in sys.path:
                normalized = path.replace('/', os.sep)
                sys.path.insert(0, normalized)
                RideLogMessage(u'Inserted \'{0}\' to sys.path.'
                               .format(normalized)).publish()`

sys.path[0] == u'[path3]'
sys.path[0] == u'[path2]'
sys.path[0] == u'[path1]'

We mock some scripts for testing, the robot libraries can't be imported and keyword autocompletion is broken, but tests work because they use the right pythonpath.


HelioGuilherme66:
I will use this PR on my fork. Thanks!

gaal-dev:
2 HelioGuilherme66

Take today version. I was busy today and made a little mistyping because it's a copy of my working version. (I put RIDE to a Docker container because RIDE still uses wxPython 2.8 and replaced namespace.py there. It works fine now:))

Issue #1845, opened 2019-03-01
Grid editor issues on new RIDE


Soothsilver:
I noticed the following differences from RIDE 1.5.2.1:

- When you hold Ctrl over a keyword, it no longer shows quick documentation.
- You cannot click the quick documentation window to have it persist (and potentially expand) while mousing over a keyword.
- The default width of column is smaller.
- When you change the width of a column, the width returns back to its automatic size as soon as you perform another action. Previously, it stayed until you switched to another test case.
- When you change the width of columns in a way that the text now fits one line, the grid row still takes up more space than a standard row if previously it was multiline. On RIDE 1.5.2.1, the grid row height adjusted.

PR #1846, opened 2019-03-04
Grid editor fixes

2Eagle2:
some fixes on grid editor:
- content assist when cancelled by ESC return correct value
- when content assist is invoked the existing contents is not cleared
- holding ctrl and hover over keyword show its information (the ctrl + M keyword is also working for backward compatibility )
- tooltips is no longer shown on other (wrong) tabs
- some idle event crashes fix.


2Eagle2:
it fix a few bugs described in issue #1845

HelioGuilherme66:
Thanks!

Issue #1819, opened 2019-02-04
Column sizing on Mac doesn't work.


technigirl:
I don't know if this a "new" issue or not, as this is my first time using RIDE on Mac, but the column sizing doesn't appear to work.  This makes the word wrap bothersome, mainly because the column size is not wide enough.  The wrap occurs before the end of the text is reached, and when I try to manually adjust the column size, the column size returns back to the automatic size when I save.  See attached picture.

I am using Mac OSX (Mojave), and Ride 1.74dev on Python3.7.
![screen shot 2019-02-04 at 3 45 33 pm](https://user-images.githubusercontent.com/6920990/52239636-8e595200-2894-11e9-9db8-8d2250e2f1b8.png)

The issue occurs even if I uncheck the "Autosize columns" box in Preferences.  Manually adjusting the default column width in Preferences also doesn't help.

technigirl:
I can work around this by unchecking Autosize in Preferences, so I don't think it's really a major issue.

technigirl:
Yes, you can install it, it is just the pip install that is broken.  You have to follow these steps:

pip install -U -r https://raw.githubusercontent.com/robotframework/RIDE/master/requirements.txt
<download code from here>
python setup.py build
python setup.py install

qupan:
Thank you,I'll try it tonight.

Nyral:
@technigirl Can you try this again with latest release? Wrapping should work differently.

technigirl:
@Nyral the word wrap and spacing problems that I reported here are resolved with the latest release.  Closing.

Nyral:
@technigirl Thank you for checking!

PR #1848, opened 2019-03-09
fix cells size in Grid editor

2Eagle2:
created cell renderer for managing size of cells in grid editor.
The values in preferences/grid now apply

Please check if the proposed behaviour is according your expectation


2Eagle2:
There is also bug in setting values in grid/preferrences - see https://github.com/wxWidgets/Phoenix/issues/1179

2Eagle2:
this patch fixes issues reported in #1845 #1819

Issue #1739, opened 2018-03-08
I'm unable to select text in cell 


MarcinJan:
I change my computer it's now running Windows 10. I don't know if Windows 10 is a reason but I have very annoying problem. I'm unable to partly select text in cell during editing. Can you help. No problem on my old computer with the same versions (Python, wxPython and ride)
![capture](https://user-images.githubusercontent.com/16239169/37140717-0440f998-22b3-11e8-8244-bb19e742dba3.PNG)

HelioGuilherme66:
@MarcinJan This problem only happens in Windows 10, with *all* versions of wxPython including latest 4.0.2 (alpha). This problem existed on Linux but it was fixed there (on my fork).
This is really a problem with RIDE and even on my fork's development branch is present. If I fix it I'll announce here.

hyh162213:
> > @MarcinJan This problem only happens in Windows 10, with _all_ versions of wxPython including latest 4.0.2 (alpha). This problem existed on Linux but it was fixed there (on my fork).
> > This is really a problem with RIDE and even on my fork's development branch is present. If I fix it I'll announce here.
> 
> May I ask a little question about the main menu of ride?
> I used ride 1.5.2.1&python2,the very old version,and now I change to ride 1.7.3.1&python3.
> When I change the ride version,I find the default width project tree on the left much smaller;
> and the grid of the keywords place much mini to input words.
> But I see some other user grid cell size looks normally,I want to change the grid size just like the pic above,any ideas please.
> My Environment is win10,python3.7.2,wxpython 4.0.4,robotframework 3.1.1,ride 1.7.3.1
> ![image](https://user-images.githubusercontent.com/46335731/51821949-b6d1d100-2315-11e9-93fe-23053eac8b14.png)

especially every time I click "save",the size of the grid I pulled bigger then get small again,it is so annoyed for me.

alepo239:
Does anyone know if this is a base issue with all of wxPython on Windows 10 or a RIDE problem?

HelioGuilherme66:
@alepo239 This is most likely RIDE's problem. You can confirm that the demo for wxPython, Custom Grid (download demo package and try), the selection of text in cells is visible.

Nyral:
This is RIDE problem. wxPython demo works fine on Windows 10.

HelioGuilherme66:
@MarcinJan You can see that fixed in current development version:
```
pip install -U https://github.com/robotframework/RIDE/archive/master.zip
```

[...]

PR #1897, opened 2019-05-05
 Fix for ticket #1739

Nyral:
I commented the function that was causing the text selection issue on windows 10.

I could have added a platform check, to skip the function only on windows 10, but I didn't notice any difference in behavior between windows 7 and windows 10. So my guess is that it's not really needed.

Expected behavior: After double clicking on the text inside a cell, the entire text should get selected. Typing characters should overwrite the text. Trying to select only a part of the text should also work.
Please check this behavior again on your side (on linux/win7/win10).


HelioGuilherme66:
There is no interference on Linux.

Issue #1857, opened 2019-03-28
Duplicate save button


zerazeon:
![DuplicateSaveButton](https://user-images.githubusercontent.com/48998198/55130604-113cb180-514e-11e9-9df6-d6595f341a6b.png)

RIDE v1.7.3.1

HelioGuilherme66:
Fixed by #1902

PR #1902, opened 2019-05-09
Fix issue  #1857: Duplicate save button

Nyral:
Removed duplicate code.


HelioGuilherme66:
I know we don't have a template for PRs, but you should know that is better to create a branch for the fix and PR, because it is easier to manage in merges to master.
Thanks.

Nyral:
Okay, got it.

Issue #1888, opened 2019-04-30
Tags events aren't working properly


Nyral:
For example, if I select a tag to edit and then select another one. The previous tag remains selected.

![image](https://user-images.githubusercontent.com/39315668/56958609-64da5a80-6b53-11e9-8cf8-6be4527a8c10.png)

The issue is caused by the OnKillFocus() function.
Originally from here: https://github.com/HelioGuilherme66/RIDE/issues/51

Nyral:
fixed and merged

Nyral:
After this fix, if you try to delete a tag on Windows 10 then RIDE will crash. I'm reopening this ticket in order to rethink the solution for this issue.

Nyral:
[More info here](https://wiki.wxpython.org/Surviving%20with%20wxEVT%20KILL%20FOCUS%20under%20Microsoft%20Windows) about why this crash was occuring.

PR #1908, opened 2019-05-12
Alternate fix for issue #1888

Nyral:
This solution should fix the issue for all platforms.
It will also not cause any crashes.


HelioGuilherme66:
Still got a crash :(

Issue #1909, opened 2019-05-13
RIDE does not allow to create .resource resource files extension


himperion92:
Since robotframework-3.1.1 you can use new ".resource" extension but RIDE does not allow to create this type of file netiher to import it as an external resource. Release notes: https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-3.1rc2.rst#new-resource-extension-for-resource-files

This issue has been reproduced in robotframework-ride-1.7.3.1 running on Python 2.7.16.

HelioGuilherme66:
Fixed by #1921

PR #1921, opened 2019-05-19
Fixes issue #1909: Added support for Resource filetype

Nyral:
It's ugly, but it works. I think... please test it.

There's also some unrelated changes. I removed a few duplicate functions from filecontrollers.

Resource option will only appear if you try to `Change Format` of resources or create a `New Resource`


HelioGuilherme66:
First tests:
I have added a normal .robot resource file to a test suite,
then I have changed its type to .resource,
- we observe, C++ complaint on content assist (we should disable it on the change type)
- the new .resource is marked in red, and name is gray in tree

image

Nyral:
After investigating this, I conclude this is a separate issue (existing from before this PR).
As such, please merge this PR as it is and create a separate ticket for this issue.

Issue #1873, opened 2019-04-07
Please bring back tag wrapping


Nyral:
[This commit](https://github.com/robotframework/RIDE/commit/c2b6c65ae5e5b0c914d48c8a2f3e67e0915f7028#diff-080617b7b2d664f4a77ee560725d0009) has removed the tag wrapping functionality. As such, tags that exceed window width are not displayed.

HelioGuilherme66:
@Nyral Please add a screenshot of an example. Thanks.

Nyral:
Here you go:

![tag wrap](https://user-images.githubusercontent.com/39315668/55926931-e9545000-5c1b-11e9-852a-d1df97477f86.jpg)

Thanks for looking into it!

HelioGuilherme66:
Fixed by #1929

PR #1927, opened 2019-05-26
Fix for issue #1873: Added tag wrapping

Nyral:
It's not 100% bug free, but at least it works exactly as before. As in, it also has the same bugs from before ;)

Bug details below:
1) Collapsible Panel does not resize correctly for the first suite, keyword,etc UNLESS the top item in the tree already has its settings panel expanded when RIDE is first started.
2) Tags do not wrap correctly when you Maximize or Minimize the Main Window (Note: This is only when there are enough tags to go on more than one row). 

Switching between test cases or closing and expanding settings panel will force the UI to refresh correctly. As such, I do not consider them to be so bad. Alternatively, we can give up on wrapping tags and add a horizontal scrollbar. This doesn't get rid of bug 1), but makes it almost unnoticeable.

PS: I moved tags at the bottom of the settings in TestCaseController and UserKeywordController in order to keep their position consistent with the other controllers.


HelioGuilherme66:
Closing in favor of #1929 
(this implementation is more complex)
Thanks @Nyral for making several options to test.

Issue #1912, opened 2019-05-15
On python3 there is no detection of file changes outside RIDE


HelioGuilherme66:
If using RIDE from Python 2.7, then the detection of a test suite having been modified by an external editor, works OK. It would cause a dialog asking to reload file from disk or discard external changes, overriding them.

This is not happening when RIDE is started from Python 3. Observed in Linux with Python 3.6.8 or Windows with Python 3.7.2.

HelioGuilherme66:
Fixed by #1928

PR #1928, opened 2019-05-28
Fixes issue #1912: Metaclass compatibility

Nyral:
The issue was present only on python 3. It was occurring because implemetation for metaclass was changed in python 3. As such, a special solution was needed to be both python 2 and python 3 compatible.


HelioGuilherme66:
There is an interesting positive side effect with this PR.
Time to run unit tests decreases a lot (50%):
```
# before this PR
$ time invoke test
(...)
real    0m32,969s
user    0m11,525s
sys     0m0,811s

# after this PR
$ time invoke test
(...)
real    0m13,952s
user    0m12,000s
sys     0m0,667s

```

HelioGuilherme66:
With this PR file change detection is OK. But Grid Editor needs to be refreshed to get changes. Text Editor only updates when reloading file. These may be done in other PR.

HelioGuilherme66:
Fixes #1912

Issue #1905, opened 2019-05-10
Add customizable colors for both Run and Text Edit in preferences


Nyral:
This includes background color, as well as text color (if possible).

PR #1941, opened 2019-06-06
Made some improvements to fix from issue #1905

Nyral:
I hope I didn't break anything...


HelioGuilherme66:
Set to *pending* because needs changes to Font Type not to be `Courier New`.

PR #1946, opened 2019-06-10
Conditioned sizes of Tagboxes and ComboBoxes to be platform specific

Nyral:
On linux (with gtk3) some sizes needed to be bigger.

In PR #1905 I made a factor related to this smaller, so on linux not all text was being shown.


Nyral:
redoing, forgot to update master

Nyral:
to be recreated

PR #1959, opened 2019-06-26
Fixes #1958: Modified OnTreeItemCollapsing to be less recursive

### LINKED ISSUE #1958: Freeze/loading when collapsing tree
Some weird loading is occurring when collapsing directories in Tree.

RIDE 1.7.4a1 (and older)
Windows 10
Python 3.7

Nyral:
Modified OnTreeItemCollapsing to be less recursive.

Fixes #1958

### DIFF
--- a/src/robotide/ui/tree.py
+++ b/src/robotide/ui/tree.py
@@ -724,30 +724,19 @@ class Tree(with_metaclass(classmaker(), treemixin.DragAndDrop,
         if node.IsOk():
             self._render_children(node)
 
-    # TODO: Remove method if CustomTreeItem defines it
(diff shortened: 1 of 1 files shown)


HelioGuilherme66:
Seems like a good idea, but It is not working. When we have animated icon, running or paused, if we collapse tree the icon remains on screen. This does not happens always.
Tested on Linux with wxPython 4.0.6.

@Nyral What you think about leaving it like it is?

Nyral:
I can confirm this is also on Windows.

Nyral:
The issue you mentioned has been fixed.

PR #1977, opened 2019-08-12
New Parser Log tab to avoid dialog when loading Test Suite

HelioGuilherme66:
There is no dialog when there are deprecated file formats or errors when loading Test Suites.


Nyral:
RIDE log is reporting the following error:
image

I assume it's because of this test:
image

Issue #1878, opened 2019-04-16
Keyword suggestion on updated RIDE (1.7.3.1) working by using 'Alt + Space' instead of 'Ctrl + Space'


imranreee:
Keyword suggestion on updated RIDE (1.7.3.1) working by using **'Alt + Space'** instead of **'Ctrl + Space'** on windows 10, at the same time it's also showing the toolbar suggestion(shown in image, please have a look).

**Note: Should work one function at a time by using 'Alt + Space'** 

![Untitled](https://user-images.githubusercontent.com/19637476/56195657-35a5f280-6057-11e9-9814-b00eec6c4499.png)

HelioGuilherme66:
The thing is that Linux, MacOS and Windows have different combinations to call the keyword help.
(For lack of space I did not fully documented this in the Shortcut Keys help)
For Windows now is: Ctrl+Alt+Space
For Linux now is: Ctrl+Space                 (Alt+Space opens command search window)
For MacOS now is: Alt+Space                 (Command+Space opens command search window)

imranreee:
Thank you for your quick reply and good to know that you have changed the shortcut functionality.  So, think should remove the Keyword Suggestion form **Alt + Space**, otherwise user will be confused like me.

kiccsepi:
Please also note, that keyword tooltips no longer appear in setup and teardown settings.
(Entering to the user keyword definition by <ctrl + click> still works.)

Nyral:
I'll see what I can do.

Nyral:
Ok, so it seems that `CmdDown()` does not work when RIDE is started with `pythonw.exe`.

Nyral:
@imranreee Ok, so it's not working because `Ctrl-Space` is now also registered for `Text Edit` and it seems to be conflicting. As a temporary solution, you can disable `Text Edit` if you want it to work.

PR #2005, opened 2019-09-14
#1878 Fixes conflict generated by TextEditor shortcuts

Nyral:
I've set it so that `TextEditor` shortcuts are unregistered if tab is not selected.

I've also fixed a small issue that I noticed in kweditor.


HelioGuilherme66:
Fixes #1878

Issue #2009, opened 2019-09-15
Executed tests get removed from tree when moving them up/down


Nyral:
Tests are not deleted, you can still see them if you go to text edit.

Steps to reproduce:

1. Select a test from a suite (pick one in the middle) and run it
2. Wait for test to finish (and icon to change to passed/fail)
3. Select test and try to move it up and down in the suite (ctrl-up/ctrl-down)
4. Test is removed

RIDE version: 1.7.4rc1.dev
wxPython version: 4.0.6
OS: Windows 10, Fedora 28

PR #2014, opened 2019-09-18
#2009 Fixes tests getting remove from tree when moving them

Nyral:
When `_set_icon_from_execution_results()` is called it sets item window to `AnimationCtrl()`, but it isn't deleted after `AnimationCtrl()` is destroyed. This was causing an error when moving test in tree.


HelioGuilherme66:
Fixes #2009

PR #2026, opened 2019-10-16
Cursor pos

### LINKED ISSUE #1913: Two edit issues
ride:1.7.4a1
os:win10 1905

issue 1:
gride editor autosize function only work if the first line is comment;
work
图片
not work
图片

issue 2:
after save in text editor,the window scroll back to top,hope it can stay at cursor's last place.

HelioGuilherme66:
Fixes #1913 part:

> issue 2:
> after save in text editor,the window scroll back to top,hope it can stay at cursor's last place.

Not fully fixed :(
Cursor still jumps to top left, but if pressing another key like Ctrl, cursor returns to position.
Probably, opening a new test suite, the cursor goes to old position

### DIFF
--- a/src/robotide/controller/ctrlcommands.py
+++ b/src/robotide/controller/ctrlcommands.py
@@ -1178,9 +1178,21 @@ class MoveRowsDown(_StepsChangingCommand):
         number_of_steps_before = len(context.steps)
         for row in reversed(self._rows):
             keep_indent = (context.steps[row].as_list()[0] == 'END' and context.steps[row-1].as_list()[0] == '')
+            remove_indent = (context.steps[row].as_list()[0] == 'FOR' and context.steps[row+1].as_list()[1] == '')
+            if context.steps[row].as_list()[0] == 'FOR':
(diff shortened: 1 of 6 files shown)


HelioGuilherme66:
@Nyral This PR contains, changes to Grid Editor and its unit tests, because I detected annoying bug when we MoveUp or MoveDown (Alt-Arrow Up/Down) rows on tests containing FOR/END blocks. That part is now fixed (although code is not the nicest).

For the main goal of the PR, keeping the cursor at the same position when Applying or Saving. I could not identify why it goes to (0,0), also Text Editor is not always in edit mode when selected. Some improvements were done to make the line with cursor visible. You can see the strange behavior of when saving the file, cursor goes to (0,0), but then if you move to other tab (or even out of focus on RIDE) and back to Text Edit, the cursor position is the correct.

HelioGuilherme66:
Another bug is on my VM with MacOS Sierra, Python 3.7.3, wxPython 6.0.6, the actions Alt-Up and Alt-Down, moves two rows instead of expected one :(.

HelioGuilherme66:
Excelent work by @Nyral, he fixed the blocking part, and now this PR can be merged.
Thanks!

Issue #1983, opened 2019-08-25
Fix Runner Log window Chinese and Latin encoding chars on Windows


HelioGuilherme66:
This is a split from #1892 where the major issue is the garbled output in Output window.
As it is now, that Output is OK, but not the Log window.

Actually, the output starts by being OK, but then other test makes it change. See here the text:
```
20190825 21:45:13.873 :  INFO : Slept 5 seconds
20190825 21:45:13.873 :  INFO : This is normal log: simplified Chinese: 汉语
Ending test:   Chinese çªºãâé.Chinese Tést.simplified Chinese: 汉语

Starting test: Chinese çªºãâé.Chinese Tést.Another € testº
20190825 21:45:13.873 :  INFO : ${options} = <selenium.webdriver.chrome.options.Options object at 0x0000027E2F48FDC8>
20190825 21:45:13.873 :  INFO : Creating an instance of the Chrome WebDriver.
20190825 21:45:17.369 :  INFO : &{mobile_emulation} = { deviceName=Galaxy S5 }
20190825 21:45:17.369 :  INFO : Creating an instance of the Chrome WebDriver.
20190825 21:45:19.953 :  INFO : Opening url 'https://www.google.com'
Ending test:   Chinese 窺䣩.Chinese T괴.Another  testڀ
```
We can see the test name is correctly logged at start (`Starting test: Chinese çªºãâé.Chinese Tést.Another € testº`, but then at end the characters are modified and originally shown as hexadecimal.

This is really a corner case of mixed encodings, and as it shows badly on the Log window but not on the Log.html, we leave this for someday :).

spitamatus:
Hello. I am running

Windows 7
Python 3.6.8
RIDE 1.7.4.2

and the problem still persists for me. The output in both output windows is showing latin chars incorrectly . The possible fixes I found as solution are:

1. #2053 
2. in #2170 as mentioned by @gogily

Are you planning on fixing this?

HelioGuilherme66:
> The output in both output windows is showing latin chars incorrectly

@spitamatus Please help us to understand what would be the difference on your system.
Please create and attach a small runnable test suite that we can run on our systems (I still have a Windows 7 around ;) ).
Run it and capture a screenshot of RIDE with both windows.
Paste here, the output of (from a command window):
`chcp`
`python --version`
`python -m pip list`
`echo %PYTHONIOENCODING%`

spitamatus:
[...]

PR #2053, opened 2019-11-17
Fixed encodings for testrunner

Nyral:
Fixed encodings in _AppendText for issue #1983

Some notes:
- Message log text is now encoded in `UTF-8`, so it should be able to display any character (if the font supports it)
- Console log will show question marks if the characters are not the same as the system encoding. I can't fix this, because they're replaced by RobotFramework. RIDE reads stdout generated by RF for console

### DIFF
--- a/src/robotide/contrib/testrunner/testrunnerplugin.py
+++ b/src/robotide/contrib/testrunner/testrunnerplugin.py
@@ -547,31 +547,14 @@ class TestRunnerPlugin(Plugin):
                           textctrl.LinesOnScreen() - 1
 
         textctrl.SetReadOnly(False)
-        try:
-            if enc:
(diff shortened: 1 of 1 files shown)


Nyral:
emoji fix moved to [separate pull request](https://github.com/robotframework/RIDE/pull/2065)

HelioGuilherme66:
@Nyral This small change in code led to many problems with Python 2.7 on Windows 7. For this, I'll remove from 1.7.4 scope, and then we will not care about python 2.7.

HelioGuilherme66:
@Nyral I updated my branch to current master and many changes, so I'll merge that one instead.

HelioGuilherme66:
Changes  merged

Issue #2039, opened 2019-10-31
Can not see grid in Edit Settings - test case


nenadfemic:
just see a little button settings in test case, but nothing load below with rows 1,2,3...
Use Python 3.8, and robotframework 3.1.2,RIDE version 1.7.3.1
![image](https://user-images.githubusercontent.com/6898640/67975357-33cb0780-fc14-11e9-868c-eb3abd28924a.png)

HelioGuilherme66:
@nenadfemic I never tested before in python 3.8. But you should try to install from releases 1.7.4b2 or from current master.

aimorari:
> @nenadfemic I never tested before in python 3.8. But you should try to install from releases 1.7.4b2 or from current master.

Unfortunately, but it does NOT help :-(

Nyral:
Look in Tools > View RIDE log and tell us if you see any error, otherwise start RIDE from command line and check there.

Also run `pip list` in command line and give us the output.

nenadfemic:
> 
> 
> Look in Tools > View RIDE log and tell us if you see any error, otherwise start RIDE from command line and check there.
> 
> Also run `pip list` in command line and give us the output.
Hi Nyral

log:

20191104 11:24:02.435 [WARN]: Importing test library "XvfbRobot" failed

Traceback (most recent call last):
Importing test library 'XvfbRobot' failed: ModuleNotFoundError: No module named 'XvfbRobot'
Traceback (most recent call last):
  File "c:\users\luka-emilija\appdata\local\programs\python\python38-32\lib\site-packages\robotide\lib\robot\utils\importer.py", line 153, in _import
    return __import__(name, fromlist=fromlist)
PYTHONPATH:
  C:\Users\luka-emilija\AppData\Local\Programs\Python\Python38-32\Scripts
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\python38.zip
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\DLLs
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\lib
  c:\users\luka-emilija\appdata\local\programs\python\python38-32
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\lib\site-packages
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\lib\site-packages\win32
  c:\users\luka-emilija\appdata\local\programs\python\python38-32\lib\site-packages\win32\lib
[...]

PR #2054, opened 2019-11-18
Fixed runtime error

Nyral:
When starting RIDE this error appears:
`RuntimeError: wrapped C/C++ object of type TagsDisplay has been deleted`

It was first mentioned [here](https://github.com/robotframework/RIDE/issues/2039#issuecomment-550014579) in issue #2039

This issue was introduced by commit from PR #1977. There, I also mentioned another issue, related to charmap encoding. I've fixed that too.

### DIFF
--- a/src/robotide/parserlog/parserlog.py
+++ b/src/robotide/parserlog/parserlog.py
@@ -125,7 +125,6 @@ class _LogWindow(wx.Panel):
 
     def _add_to_notebook(self, notebook):
         notebook.add_tab(self, 'Parser Log', allow_closing=True)
-        notebook.show_tab(self)
         self._output.SetSize(self.Size)
 
     def close(self, notebook):
--- a/src/robotide/parserlog/parserlog.py
+++ b/src/robotide/parserlog/parserlog.py
@@ -19,12 +19,6 @@ import uuid
 import atexit
 import glob
 import sys
-try:
-    from wx.lib.agw.aui.auibook import AuiNotebook as auib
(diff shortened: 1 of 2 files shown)


src/robotide/parserlog/parserlog.py
```diff
@@ -19,12 +19,6 @@
 import atexit
 import glob
 import sys
-try:
```
HelioGuilherme66: This import is important to be checked if in wxPython < 3.0.3. We should keep it for now.

Nyral: I understand, but the import is not used in this file. There is no point in keeping it.

HelioGuilherme66: OK. I think I left it there for future use.

src/robotide/parserlog/parserlog.py
```diff
@@ -125,7 +119,6 @@ def __init__(self, notebook, log):
 
     def _add_to_notebook(self, notebook):
         notebook.add_tab(self, 'Parser Log', allow_closing=True)
-        notebook.show_tab(self)
```
HelioGuilherme66: This is needed to show Parser Tab when selected from Menu.

Nyral: Yeah, but that's what was actually causing the error. It occurs when it tries to switch to Parser Tab at startup. I guess I could move it... to avoid error and show Parser Tab when selected from Menu.

Issue #2027, opened 2019-10-17
Separate Font and Colorization


vyvital:
Looks like Font and Colorizations are in the same function and both depend on the `Pygments` library thus both get skipped if the library is missing.
Moreover when uninstalling the library the same font will look differently.

Also it looks like there is an issue when swapping between Courier and Courier New.
When swapping it doesnt do anything, but when picking another font it does.
Maybe because these 2 are the only fonts that contain the same word?

`1.7.4b2`

HelioGuilherme66:
Fixed by #2056

PR #2056, opened 2019-11-18
Separated colorization and font dependency on Pygments in Text Editor

Nyral:
Fixes issue #2027

Additionally, I've also modified preferences dialog for Text Editor so that unavailable colorization settings are not added when `Pygments` is not installed.


Nyral:
thanks

PR #2057, opened 2019-11-19
Fixed colorization error in large test cases

Nyral:
I reverted `_coloring_task` and `colorize` to an older implementation where recursion error doesn't occur. More specifically, to [before this commit](https://github.com/robotframework/RIDE/commit/de18d56457490ef30857c0efc26ca5e4d1133766).

I haven't seen any issues, but I might have missed something. Please pay attention when testing this.


HelioGuilherme66:
To have a test to validate the bug and the fix, we can generate with:
```
rfgen.py -s 1 -t 1 -e 1000
```

PR #2064, opened 2019-11-26
Added zoom factor to settings.cfg so that changes persist in Text Editor

### LINKED ISSUE #1298: Persist font size changes done in text editor
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=1260) by sajjadmalang on 18 Jun 2013_

The default font size on text edit tab is too big. So i ctr+scroll to reduce the font size. But when i switch to the edit tab and then back to the "text edit" tab the font size is reset to the initial size. should be this an enhancement request or bug report?

Thanks,

version
RIDE 1.1 running on Python 2.7.2. 
Robot Framework 2.7.7 (Python 2.7.2 on win32)
Win7

Nyral:
I've chosen to add zoom factor as a separate input in Text Editor preferences. The original issue mentioned that it should be added to font size, but I thought that would just create confusion. This way the user can be aware of its value.

I've also made the margin resizable, so that it adapts to the new font size/zoom.

The new function `build_UnaryOp` is so that configobj can handle variable from settings.cfg that have negative values. I had to fix this too because zoom factor range is [-10, 20]. 

Fixes #1298

### DIFF
--- a/src/robotide/editor/texteditor.py
+++ b/src/robotide/editor/texteditor.py
@@ -693,7 +693,7 @@ class RobotDataEditor(stc.StyledTextCtrl):
     def calc_margin_width(self):
         style = stc.STC_STYLE_LINENUMBER
         width = self.TextWidth(style, str(self.GetLineCount()))
-        return width + self.TextWidth(style, "12")  # return with offset
+        return width + self.TextWidth(style, "1")
 
     def get_selected_or_near_text(self):
         # First get selected text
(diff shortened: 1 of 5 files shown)


Nyral:
I fixed the conflicts too.

Nyral:
done

HelioGuilherme66:
@Nyral Thanks for implementing an improvement requested on 2013!

PR #2072, opened 2019-11-30
Fixed runtime error again

Nyral:
When starting RIDE this error appears:
`RuntimeError: wrapped C/C++ object of type TagsDisplay has been deleted`

It was first mentioned here in issue #2039

Error occurs because RIDE tries to show Parser Log tab at startup.
To fix it I added a second condition to `OnViewLog()`, so that Parser Log tab is only shown when the user selects from Menu.

The `show_tab` call at line 122 should remain removed.


Nyral:
> 
> 
> Good solution.
> (it is strange that the code for RIDE log was equal to Parser Log and the problem did not happen there]

It's because you set Parser Log to open the tab at startup if there are errors. RIDE log tab can only be opened from the menu.

PR #2076, opened 2019-12-01
Condition for opening with encoding on Python 2.7

HelioGuilherme66:


src/robotide/parserlog/parserlog.py
```diff
@@ -61,7 +63,10 @@ def _remove_old_log_files(self):
     @property
     def _logfile(self):
         if self._outfile is None:
-            self._outfile = open(self._path, 'w', encoding='utf8')
+            if PY2:
+                self._outfile = io.open(self._path, 'w', encoding='utf8')
+            else:
+                self._outfile = open(self._path, 'w', encoding='utf8')
```
Nyral: There is no need for condition. Use io.open for both python 2 and python 3. 
On python 3 io.open is the same as open. Only on python 2 it's different.

src/robotide/log/log.py
```diff
@@ -20,10 +20,13 @@
 import atexit
 import glob
 import sys
+import io
 
 from robotide.pluginapi import Plugin, ActionInfo, RideLog
 from robotide import widgets
 from robotide import context
+from robotide.context import IS_WINDOWS
```
Nyral: this isn't used?

src/robotide/log/log.py
```diff
@@ -62,7 +65,10 @@ def _remove_old_log_files(self):
     @property
     def _logfile(self):
         if self._outfile is None:
-            self._outfile = open(self._path, 'w', encoding='utf8')
+            if PY2:
+                self._outfile = io.open(self._path, 'w', encoding='utf8')
+            else:
+                self._outfile = open(self._path, 'w', encoding='utf8')
```
Nyral: There is no need for condition. Use io.open for both python 2 and python 3.
On python 3 io.open is the same as open. Only on python 2 it's different.

src/robotide/parserlog/parserlog.py
```diff
@@ -19,11 +19,13 @@
 import atexit
 import glob
 import sys
+import io
 
 from robotide.pluginapi import Plugin, ActionInfo, RideParserLogMessage
 from robotide import widgets
 from robotide import context
-
+from robotide.context import IS_WINDOWS
```
Nyral: this isn't used?

src/robotide/contrib/testrunner/runprofiles.py
```diff
@@ -158,10 +158,12 @@ def _get_arguments(self):
     def _parse_windows_command(self):
         from subprocess import Popen, PIPE
         try:
+            # print("DEBUG: parser_win_comm Enter arguments: %s" % self.arguments)
             p = Popen(['echo', self.arguments], stdin=PIPE, stdout=PIPE, stderr=PIPE, shell=True)
             output, _ = p.communicate()
             output = str(output).lstrip("b\'").strip()
-            self.arguments = output.replace('"', '').replace('\'', '').replace('\\r\\n', '')
+            self.arguments = output.replace('"', '').replace('\'', '').replace('\\\\', '\\').replace('\\r\\n', '')
```
Nyral: Why did you add this?

src/robotide/log/log.py
HelioGuilherme66: Correct. I was trying other options in Windows.

src/robotide/log/log.py
HelioGuilherme66: OK.

Issue #1687, opened 2017-02-28
RIDE Console  Output Size Limit?


Farb:
I met a strange problem when I used RIDE to call a .exe program by `Process` library.And the program output some logs,then I used the `Log ` output the infos to the console of RIDE.My case is as follows:
·${dailyJob}       RYP2P.Schedule.exe jobkey:dailydatajob runParams:startdate=2017-1-01$runmode=test$userid=-1 autoclose:true  
·
```
    RunDailyJob
    [Documentation]    test Job
    OperateConfigInfo    update ry_modulesign set Online=1;
    Log Many    ensure my Job offline
    ${result}    Run Process    ${dailyJob}    shell=true    cwd=RyP2P.Schedule    stdout=${CURDIR}/stdout.txt
    Log    ${result.stdout}
```

The .exe program will execute a very complicated business logic.For example,it will loop through a year,the step length is one day.In every day,the program
will outputs a words like 'today is ${today}'。Therefore,the progarm will outputs lots of content finally.In my program,the problem is that the case cannot stop.
If you write the log to the file in the program,you will find the loop will pause in some stage such as the very day of half year.However，when you click the stop button
of RIDE，the loop will continue running to the end of the program.If I comment the `Log    ${result.stdout}`,the case will end quickly.I guess that if RIDE has a limit over 
the console size or it is a bug of RIDE?

Thanks in advance!

Casalis:
Have you still got this problem? 
How many lines does the ".exe" return for each day?
Is the Log file longer that the RIDE console log?

HelioGuilherme66:
Closing because is too old and we cannot reproduce, or lack of information.

Nyral:
You can use the test in this suite to reproduce the issue. It should crash after 300 iterations or so.
[MessageLog.txt](https://github.com/robotframework/RIDE/files/3908836/MessageLog.txt)

To sum it up, we need to add character limit for each line and also maybe periodically clear memory log and console if number of lines/chars reach a certain size.

HelioGuilherme66:
@Nyral With your example, I cannot reproduce (today's development version):
![image](https://user-images.githubusercontent.com/2866630/69969290-0f649280-1514-11ea-967f-6c6471e0c4bb.png)

HelioGuilherme66:
This is fixed by stopping the log when a memory limit is reached. The following message is shown in the log:
```
Messages log exceeded 80% of process memory, stopping for now...
```

PR #2086, opened 2019-12-16
Separates AppendText for MessagesLog. Adds process memory limit on Me…

HelioGuilherme66:
…ssagesLog.

Possible fix for #1687


src/robotide/contrib/testrunner/testrunnerplugin.py
```diff
@@ -70,7 +72,7 @@
 from robotide.robotapi import LOG_LEVELS
 from robotide.utils import robottime, is_unicode, PY2
 from robotide.preferences.editors import ReadFonts
-from sys import getfilesystemencoding
+from sys import getfilesystemencoding, getsizeof
```
Nyral: `getsizeof` is not used

Nyral:
It still crashes for me.

HelioGuilherme66:
Using the memory_profile tool, we see that on the AppendText with encoding it takes about 200-500KB of memory increase. That is why I have isolated the messages_log. But I did not changed the output window. Also if you run from command line, the most of memory use is in the end when processing output.xml to get log and report. Don't really know what else we can do, other then suggesting running `robot` outside of RIDE.

Nyral:
Try running the test 2-3 times ;)

Your solution prevents text from added to memory log, but python ram usage still increases.
Console log is affected too, just use `Log to console` keyword instead.

I've noticed that ram usage is freed if you uncheck `Show message log`. That's because it destroys the object. Maybe we can try that? If memory usage increases too much, then destroy and recreate the textctrl?

HelioGuilherme66:
Experimenting memory_profile (run all 1000 iteractions, but computer got slow when log/report generation):
`PYTHONPATH=/usr/local/lib/python3.7/site-packages mprof run src/robotide/__init__.py `

This is the chart generated by `mprof plot`:
image

Nyral:
With the changes in latest commit, the memory limit no longer triggers.

HelioGuilherme66:
It triggers, writes message on log, but is too fast to be noticed, because new messages are logged.

Nyral:
Did you try running the MemoryLimit test again after you made those changes? RIDE crashes.

HelioGuilherme66:
This PR is going to be on hold, because there is no noticeable advantage in merging it. Below are the charts for the test runs (3 consecutive runs with counter at 200). We can see a 500MB increase on each run. Further code changes must reduce that 500MB to a value near 0.
Before PR (master branch):
memory_before_PR

After PR (memdebug branch):
memory_with_PR

Was tested on Windows 10, python 3.7.4 wxPython 4.0.7.post2.

PR #2099, opened 2019-12-28
Fixed shortcut in keyword tooltip. We no longer use Ctrl-M.

Nyral:


Nyral:
That method would cause localize_shortcut to be executed every time you hover a keyword cell. I've updated the implementation so that it's executed only once.

PR #2102, opened 2019-12-30
Makes Text Editor read-only if file is read-only. Visual hint is back…

### LINKED ISSUE #987: Editing controls should be disabled when file status is READ ONLY
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=950) by @mkorpela on 9 Mar 2012_

see <a class="closed_ref" href="https://code.google.com/p/robotframework-ride/issues/detail?id=157" title="RIDE should not allow editing read-only files"> issue 157 </a> .

HelioGuilherme66:
…ground color change.

Not perfect, because changing background color on Preferences, when on Read-Only file, does not apply read-only. 
Correct on Tab change, Open file and select from tree.

Fixes #987

### DIFF
--- a/doc/other/example_maven_runner/pom.xml
+++ b/doc/other/example_maven_runner/pom.xml
@@ -20,7 +20,7 @@
 					<plugin>
 						<groupId>org.robotframework</groupId>
 						<artifactId>robotframework-maven-plugin</artifactId>
-						<version>1.5.1</version>
+						<version>1.5.2</version>
 						<executions>
 							<execution>
 								<phase>integration-test</phase>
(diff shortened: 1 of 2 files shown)


Nyral:
Read only files are not loaded into text editor if the tab is focused. Instead, content of last non read only file remains there. Also, background color doesn't change when white is used.

Wouldn't it be easy to just make the tab unselectable/grayed out? This change is a bit counter intuitive.

HelioGuilherme66:
> Read only files are not loaded into text editor if the tab is focused. Instead, content of last non read only file remains there. 
**Fixed**
Also, background color doesn't change when white color is used. With others, the color is shifting. 
**Fixed** Problem was with named colors instead of hex RGB.
> This change is not very intuitive.
> 
> Wouldn't it be easier to just make the tab unselectable/grayed out? (see wxpython demo for AUI)
> 
I prefer like this, because Search function is still usable.
> If you really want to modify colors for read only files, then enforce the default colors stored in settings.cfg (the one from preferences folder) and use wx.LIGHT_GREY for background. Additionally, you could also call wx.Bell when the user tries to edit a read only file.
I like the color change and now is more limited to gray on bright colors.
The Bell is so annoying in Grid editor.

PR #2113, opened 2020-01-13
Fixed SelectAll issue in Grid Editor, RIDE Log and Parser Log

### LINKED ISSUE #2084: ctrl+A doesn't do anything
On older RIDE (1.5 and older) I was able to select contents of the whole case when I pressed ctrl+A.
But on version RIDE 1.7.4rc1.dev running on Python 3.7.4 (on Windows 7) it doesn't do anything.

What's the alternative for that now?
Is it possible to bring it back?

Nyral:
Fixes #2084

Issue stems from the order in which the shortcut conditions are called in kweditor. Additionally, there is a conflict between RIDE and Parser Log tab shortcuts.

Changes:
- improved `if` case structure in `OnKeyDown`  (functionality should be the same or better)
- made RIDE and Parser Log tabs to unregister their shortcuts when they're not selected
- updated code so that `Select All` works in RIDE and Parser Log tabs (wasn't implemented)

Please check to see that shortcuts work. On MacOS there might be some additional conflicts between shortcuts from Grid Editor and Text Editor, so make sure to check shortcuts both when Text Editor is disabled and enabled. Leave a message if you find any discrepancies.

Sorry it took so long, I've been busy with other stuff and kinda forgot.

### DIFF
--- a/src/robotide/parserlog/parserlog.py
+++ b/src/robotide/parserlog/parserlog.py
@@ -100,6 +100,7 @@ class ParserLogPlugin(Plugin):
             self._panel = _LogWindow(self.notebook, self._log)
             self.notebook.SetPageTextColour(self.notebook.GetPageCount()-1, wx.Colour(255, 165, 0))
             self._panel.update_log()
+            self.register_shortcut('CtrlCmd-C', lambda e: self._panel.Copy())
         if show_tab:
             self.notebook.show_tab(self._panel)
 
@@ -142,16 +143,13 @@ class _LogWindow(wx.Panel):
     def OnKeyDown(self, event):
         keycode = event.GetKeyCode()
 
-        if event.ControlDown():
-            if keycode == ord('C'):
-                self.Copy()
-            elif keycode == ord('A'):
-                self.SelectAll()
+        if event.ControlDown() and keycode == ord('A'):
+            self.SelectAll()
         else:
             event.Skip()
 
     def Copy(self):
-        self._output.Copy()
+        pass
 
     def SelectAll(self):
         self._output.SetSelection(-1, -1)
--- a/src/robotide/parserlog/parserlog.py
+++ b/src/robotide/parserlog/parserlog.py
@@ -21,7 +21,7 @@ import glob
 import sys
 import io
 
-from robotide.pluginapi import Plugin, ActionInfo, RideParserLogMessage, RideNotebookTabChanging
+from robotide.pluginapi import Plugin, ActionInfo, RideParserLogMessage
 from robotide import widgets
 from robotide import context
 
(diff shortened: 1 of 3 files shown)


src/robotide/parserlog/parserlog.py
```diff
@@ -97,29 +99,42 @@ def _log_message(self, log_event):
...
+            self._register_shortcuts()
+        elif message.oldtab == self.title:
+            self.unregister_actions()
+
+    def is_focused(self):
+        return self.notebook.current_page_title == self.title
+
+    def _register_shortcuts(self):
+        self.register_shortcut('CtrlCmd-C', lambda e: self._panel.Copy())
+        self.register_shortcut('CtrlCmd-A', lambda e: self._panel.SelectAll())
```
HelioGuilherme66: I managed to get Parser log receiving Ctrl-A even if tab was closed. It received Ctrl-A from kweditor and text editor (not working there). This is console outuput:
```
  File "C:\github\RIDE\src\robotide\ui\actiontriggers.py", line 338, in __call__
    action.act(event)
  File "C:\github\RIDE\src\robotide\action\action.py", line 89, in act
    self.action(event)
  File "C:\github\RIDE\src\robotide\parserlog\parserlog.py", line 121, in <lambda>
    self.register_shortcut('CtrlCmd-A', lambda e: self._panel.SelectAll())
  File "C:\github\RIDE\src\robotide\parserlog\parserlog.py", line 163, in SelectAll
    self._output.SetSelection(-1, -1)
RuntimeError: wrapped C/C++ object of type TextCtrl has been deleted
```
(Windows 10 with Python 2.7.16)

Nyral: Hm, good find. It seems that shortcuts are not unregistered when tab is closed. I've tried several workarounds, but they all fell short. I think `register_shortcut` is meant to be used only when there are menu entries involved. As such, let's use key events as much as possible from now on. These cannot create conflicts because they are bound to the object inside the tab.

Nyral: I've updated RIDE and Parser Log with the new method. Please check it out.
If everything is ok then should I proceed and update Text Editor using the same method?

HelioGuilherme66: Parser log is a copy of RIDE Log. The Select All keyboard shortcut is working on both windows, but the Command-C it is not. This is not a problem because users can use the right click menu Copy function. This is just a minor problem. On MacOS we have other problems with key controls, for example, I discovered that  the Delete key is not working on Text Editor.

HelioGuilherme66: Parser log is a copy of RIDE Log. The Select All keyboard shortcut is working on both windows, but the Command-C it is not. This is not a problem because users can use the right click menu Copy function. This is just a minor problem. On MacOS we have other problems with key controls, for example, I discovered that  the Delete key is not working on Text Editor.

PR #2152, opened 2020-03-01
Turns Test Suite Tree into a Treeplugin

HelioGuilherme66:
Now the Test Suite Tree can be disabled, just like any other plugin.
With the new top menu, View it can be shown (or with F12 key).
The Plugin when disabled does not show the View menu option.
The panel has its own view control by the setting "opened", which will be True from the menu View.

### DIFF
--- a/README.rest
+++ b/README.rest
@@ -19,6 +19,13 @@ See the `release notes`_ for latest release version 1.7.4.1.
 
 **Attention, don't install RIDE on Python 3.8 (wait for release of version 2.0)**
 
+**Or you can install from development, where running on Python 3.8 is supported (since February 2020)**
+
+
+Install with:
+
+``pip install -U https://github.com/robotframework/RIDE/archive/master.zip``
+
 **See the FAQ at Wiki**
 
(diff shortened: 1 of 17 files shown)


Nyral:
A few issues: 
- [ ] Test Suite Tree appears under Files Tree and it's not possible to move it at top. Please have them appear in same positions as they were before.
- [ ] Moving Test Suite Tree to the right of the Files Tree causes RIDE to crash (possible other positions too)
- [ ] Moving any of the plugins causes black background to appear on their old positions. This isn't normal AuiManager behavior. Something is causing an error for that to happen.
- [ ] Double clicking on the Test Suite Tree top bar causes it to become duplicated, followed up by RIDE crash
- [ ] getting this error on startup
```
C:\Python37\lib\site-packages\wx\lib\agw\aui\framemanager.py:4742: UserWarning: A pane with the name 'tree_content' already exists in the manager!
```

Fix these first and I'll look

HelioGuilherme66:
I gave some explanations on Slack.

Issue #2148, opened 2020-02-27
RIDE TaskBar icon in Win10 is black on dark grey... hard to see


CMTaylor:
![RIDE_Icon](https://user-images.githubusercontent.com/1346041/75463005-29671300-5953-11ea-8165-23ba90ea05f8.png)
The "Robot Face" icon is fine, but please change it to a brighter colour (like green passing icon or blue executing icon in RIDE) so that it can be seen in Win10.

HelioGuilherme66:
@CMTaylor If you like, just replace the file [..\site-packages\robotide\widgets\robot.ico](https://github.com/robotframework/RIDE/blob/master/src/robotide/widgets/robot.ico) in your installed files.

PR #2154, opened 2020-03-02
Added white color to inside the icon face

Nyral:
Before:
image
After:
image

Should be much more visible like this. 
#2148


HelioGuilherme66:
Closes #2148

PR #2155, opened 2020-03-02
Fixed crash when you delete tag and more

### LINKED ISSUE #2146: RIDE crashes when you delete a tag via the GUI
Open RIDE and navigate to a test case with multiple tags. Highlight one of them in the GUI. Delete the text of the tag and press <ENTER>. RIDE will crash with no error message or log that shows what error caused the crash.

If you go into the TextEdit tab and delete the tag that way, then its OK. Just the direct GUI delete causes the crash.

Nyral:
I also fixed 2 additional issues I found while investigating this:
- Tagbox is left empty, without being destroyed, if you empty the content of a new tag
image
- the following runtime error appears if you click on another tag, instead of pressing ENTER key after you delete a tag
```
  File "C:\Python37\lib\site-packages\robotide\editor\tags.py", line 191, in OnKillFocus
    if self.value != '':
  File "C:\Python37\lib\site-packages\robotide\editor\tags.py", line 204, in value
    return self.GetValue().strip()
RuntimeError: wrapped C/C++ object of type TagBox has been deleted
```
Fixes #2146

### DIFF
--- a/src/robotide/editor/tags.py
+++ b/src/robotide/editor/tags.py
@@ -188,7 +188,7 @@ class TagBox(wx.TextCtrl):
         self._update_value()
         # Send skip event only if tagbox is empty and about to be destroyed
         # On some platforms this event is sent too late and causes crash
-        if self.value != '':
+        if self and self.value != '':
             event.Skip()
 
     def _update_value(self):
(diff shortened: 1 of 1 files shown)


CMTaylor:
Not having this fix in a Python 2.7 version of RIDE is a pain. I'm not going to be able to upgrade to Python 3 for several months yet, due to company priorities.

HelioGuilherme66:
@CMTaylor The fix is small and compatible with python 2.7, you can just apply the changes to your installed file.

PR #2158, opened 2020-03-10
Encloses text in Cell or selected text with certain symbols

HelioGuilherme66:
When in Cell Editor we have selected text, it will be enclosed if pressed these keys:
'{', '[', '(', "'", "'", '`'
If no text selected writes the right symbol, or a duplicate.

If in navigation mode in Grid Editor, uses the full cell content to apply the enclosing.


HelioGuilherme66:
To Reviewers, I have set this to Work In Progress, because I need to make the cursor appear inside "" or '' or ``, the other cases are OK.

PR #2198, opened 2020-05-12
Log misc fixes

### LINKED ISSUE #2197: Taking log plugin into use failed
Environment:
Windows 10 Enterprise (1909)
Python 3.8.2 x64
Robot Framework 3.2.1
Robot Framework AutoItLibrary 1.2.5
Robot Framework RIDE 2.0b1.dev1
WMI 1.5.1
wxPython 4.0.7.post2

When launching RIDE via the desktop shortcut with the command: "C:\Program Files\Python38\pythonw.exe" -c "from robotide import main; main()"

A message dialog is displayed that Taking log plugin into use failed.  If you click OK, RIDE opens.

Better information is available from the plugin manager inside RIDE.
Log error 1
Log error 2

If anybody needs more details or logging (tell me where to find it), let me know.  RIDE runs.  It's just an annoying prompt every time you launch.

HelioGuilherme66:
Attempts to fix #2197

### DIFF
--- a/src/robotide/contrib/testrunner/runprofiles.py
+++ b/src/robotide/contrib/testrunner/runprofiles.py
@@ -228,7 +228,7 @@ class PybotProfile(BaseProfile):
     def _create_error_log_message(self, error, returncode):
         # bash and zsh use return code 127 and the text `command not found`
         # In Windows, the error is `The system cannot file the file specified`
-        if 'not found' in error or returncode is 127 or \
+        if 'not found' in error or returncode == 127 or \
                         'system cannot find the file specified' in error:
             return pluginapi.RideLogMessage(
                 RF_INSTALLATION_NOT_FOUND, notify_user=True)
--- a/src/robotide/log/log.py
+++ b/src/robotide/log/log.py
@@ -57,7 +57,7 @@ class LogPlugin(Plugin):
             try:
                 os.remove(fname)
             except OSError or IOError as e:
-                sys.stderr.write("{}".format(e))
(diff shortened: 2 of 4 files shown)


CMM-Guru:
Ok.  Your comment about the log file is somewhat on track.

#1 - pythonw -m robotide.__init__ from the command line still has the problem
       python -m robotide.__init__ deletes the file and works as expected repeatedly.
       This seems to have something to do with how pythonw vs python terminates.
       If I change my shortcut to python.exe it works, except I get a command prompt every time I launch that does not close.  It appears that if I use pythonw.exe I uses what I would call a "hidden" command prompt.  The process does not actually display the command prompt window underneath RIDE.

I am the owner of the log file in the temp directory (as would everybody else be on their system).  And I have full rights  (R/W/Modify) to the file.

If I delete the file, RIDE will launch properly and create a new file.  I can exit and relaunch as many times as I want.

If I launch RIDE and open a .robot file and then save it (not even changing it) then exit RIDE, I have the problem again.

The reason the problem now exists is that although RIDE ended, Python did *not*.  And Python is still holding on to the file, so the next launch of RIDE cannot access it.  You must (without RIDE running) go to Windows Task Manager and end all Python processes.

If I open RIDE and Exit, I can do it 10 times in a row with no problems.

If I open RIDE, open and save a file, then exit RIDE, I get a copy of the Python process that did not shut down.

If I do this 10 times, I have 10 copies of the Python process running!

Hopefully this is some further useful information?

PR #2205, opened 2020-06-10
Update treenodehandlers.py to prevent freeze ups

### LINKED ISSUE #1843: System freezes on Windows 10
When using RIDE with large suite (1,000 tests), if you're using Windows 10, RIDE randomly causes the entire system to freeze for 10 seconds up to 1 minute when you click on anything in the RIDE window. I've seen this happen on many computers, all of them with Windows 10. It never happens on Windows 7.

This happens both in RIDE 1.5.2.1 (with recommended wxPython) and with RIDE 1.7.3.1 (I tested with wxPython 4.0.4, Python 2.7).

cbeytas:
The _ActionHandler class is based off wx.Window which causes a native widget to be created. Avoid adding this native window to the tree to prevent Windows OS freezes when the tree is scrolled. Add it to the tree's parent instead.
Looks like the only wx.Window feature utilized is for popup menus when a tree item is right-clicked.
Attempts to fix #1843.

### DIFF
--- a/src/robotide/ui/treenodehandlers.py
+++ b/src/robotide/ui/treenodehandlers.py
@@ -88,7 +88,7 @@ class _ActionHandler(wx.Window):
     _label_open_folder =  'Open Containing Folder'
 
     def __init__(self, controller, tree, node, settings):
-        wx.Window.__init__(self, tree)
+        wx.Window.__init__(self, tree.GetParent())
         self.controller = controller
         self._tree = tree
         self._node = node


HelioGuilherme66:
Thank you @cbeytas for the PR. In the future, please create a branch for the modifications and PR, it helps organizing code merging and history of changes. You can do this very easily after updating `master` (`git pull upstream master), with a `git checkout -b fix_1843` (in this case).

Nyral:
> If you start RIDE with Tree hidden/closed, and open a test suite folder, then when restored the Tree, right-clicking does not show the menu, and expanding test suites does not show test cases.
@HelioGuilherme66 

Based on my observations, this occurs only when RIDE starts with Tree floating or it was floating previously and it was hidden afterwards. 
The changes proposed cause all widgets to be moved to the frame holding the Tree. But if Tree is floating then they are moved to the AuiFloatingFrame, not RideFrame. As such, when the AuiFloatingFrame is closed the widgets are lost.

HelioGuilherme66:
> Based on my observations, this occurs only when RIDE starts with Tree floating or it was floating previously and it was hidden afterwards. Can you confirm?

This is a problem of Tree panel, the implementation is not the best because the tree is only created at init time. The update tree is not working well. I would not like to go back to the static position of tree panel. Best is to fix the tree update and refresh (messaging trigger).

HelioGuilherme66:
Thank you for the PR, I am going to merge because at least removed GTK error messages.

Issue #1668, opened 2016-11-24
Opening large amount of tests in RIDE folder tree freezes


Rochet2:
I have a directory that contains 800 tests.
Those tests are foldered into multiple folders that have subfolders and so on.
If I open multiple folders / suites that contain tests at the same time, RIDE starts to freeze the whole computer.

I did not seem to notice any change in RAM or CPU usage.
The freezing happens so that when you are in RIDE and click on something - anything: a button in RIDE, some other program on task bar - anything.
The freezing can take from some seconds to for example 10 seconds or possibly more. Usually it is a relatively long time.
Once the freezing happens once, it happens from that  point onwards until RIDE is closed and opened again.

You can try to reproduce this by having a directory that contains 5 folders.
Those folders contain 10 suites each.
The suites contain 1 to 10 tests each.
(Ok, lets just say that the whole left folder tree in RIDE is full of suites and tests. Completely white - having 40 suites open or similar)
When you open all this up in RIDE it will start to freeze. It may require a bit of waiting or fiddling around or trying to click on RIDE and some other program and then RIDE again multiple times or clicking something in RIDE to get the freezing to start happening.

It seems that I am not the only one having this problem as I have witnessed others having it as well.

We all used the latest release, which is https://github.com/robotframework/RIDE/releases/tag/v1.5.2.1

RIDE is a nice tool compared to maintaining things manually in the text files - but this freezing keeps interrupting work and forces restarting RIDE multiple times every day.

With the folder tree I mean the left side of RIDE
![ride_1](https://cloud.githubusercontent.com/assets/468816/20605254/1aa926e4-b273-11e6-90ba-0f5ccbe00dd7.png)

Rochet2:
[tests.zip](https://github.com/robotframework/RIDE/files/1016451/tests.zip)
Here is an example of some test data you can use to try freeze your RIDE.
It contains 160 folders that have 10 suits that have 10 tests each.

To use the data simply place it somewhere, open RIDE, click to open directory in RIDE and select the Tests folder. To test the freeze simply scroll the tree view up and down once by grabbing the scroll bar with your mouse and then try to change the window focus to any other program, like notepad for example. The system should freeze and you cannot do anything.

I tried profiling RIDE and disabling everything that seemed like a problem, but the freeze still exists.
I disabled for example _sendMessage in publisher entirely as it was called 100 000 times.

The profiling would suggest that the whole program time is almost entirely (not counting data loading) spent in the GUI MainLoop
[...]

PR #2210, opened 2020-06-16
Fix for 1668: Given a large test suite, selecting all tests from the root of tree, freezes RIDE for some time

valeriobruno:
1668: Selecting and deselecting tests using the right-click commands on the root of a deeply nested test folder, freezes RIDE for several minutes.
    The problem is caused by the strategy to find the test to select/deselect: expanding and collapsing all the nodes in the Tree plugin. This causes CPU waste to expand/collapse (and maybe render) and  create the tree node handlers.
    The idea of the fix is to navigate the tests using the link between controllers, in order to avoid all the dancing of the tree nodes.

To evaluate the improvement: open a large test suite. right-click on the root of the tree and "Select all tests". Click on another node of the tree. Observe the time required for the UI to return responsive. This should be noticeable even without a precise measurement. In fact, in my environment it passed _from minutes to seconds_.

A large test suite for testing is available in this comment: https://github.com/robotframework/RIDE/issues/1668#issuecomment-640885263

Notes: this code change has no effect on the initial load time of the suite. Nor affects the other performance issues of the Tree plugin. In any case, it's a relevant improvement.

### DIFF
--- a/src/robotide/contrib/testrunner/testrunnerplugin.py
+++ b/src/robotide/contrib/testrunner/testrunnerplugin.py
@@ -381,7 +381,7 @@ class TestRunnerPlugin(Plugin):
         return ret == wx.YES
 
     def _tests_selected(self):
-        return len(self._names_to_run) != 0
+        return len(self._selected_tests) != 0
 
     def ask_user_to_run_anyway(self):
         ret = wx.MessageBox('No tests selected. \n'
--- a/src/robotide/contrib/testrunner/testrunnerplugin.py
+++ b/src/robotide/contrib/testrunner/testrunnerplugin.py
@@ -137,7 +137,11 @@ class TestRunnerPlugin(Plugin):
         self._test_runner = TestRunner(application.model)
         self._register_shortcuts()
         self._min_log_level_number = LOG_LEVELS['INFO']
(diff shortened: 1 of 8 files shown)


Nyral:
I'm getting the following error:
```
  File "C:\Python37\lib\site-packages\robotide\utils\eventhandler.py", line 24, in wrapped
    if self and self._can_be_edited(event):
  File "C:\Python37\lib\site-packages\robotide\utils\eventhandler.py", line 42, in _can_be_edited
    ctrl = self.get_selected_datafile_controller()
  File "C:\Python37\lib\site-packages\robotide\ui\treeplugin.py", line 676, in get_selected_datafile_controller
    dfnode = self._get_selected_datafile_node()
RecursionError: maximum recursion depth exceeded
```

valeriobruno:
> Also, RIDE crashes if select/deselect all tests is done after expanding one suite. Crash is due to RecursionError, so probably same root cause.

I've reproduced the same error with your test, although my stack was different.
I've delayed the processing of the RideTestSelectedForRunningChanged event to the next WX loop.

Can you check now?

Nyral:
Looks like that fixed it. Good job.

PR #2232, opened 2020-07-12
Redesign RideEventHandler and file system monitoring feature

jnhyperion:
- Directory workspace
In the following scenario will trigger reload workspace confirm dialog, when confirmed, it will reload whole workspace, otherwise, it will discard changes
1. create/modify/delete/rename files from outer editor, only works for .robot,.txt,.tsv,.resource
2. create/rename/delete sub folders
3. rename/delete workspace root folder, note: in the case, when confirmed reload, will ask user to open new directory

- Single suite file workspace
if this file is renamed/deleted, will also trigger reload workspace confirm dialog, when confirmed reload, will ask user to open new directory

Tested in windows, macOS and Ubuntu with wxPython 4.1.0.

**Looks like wx.FileSystemWatcher cannot work properly in MacOS with wxPython 4.0.7.post2**

### DIFF
--- a/src/robotide/ui/mainframe.py
+++ b/src/robotide/ui/mainframe.py
@@ -602,9 +602,9 @@ class RideFrame(wx.Frame):
                             style=wx.YES_NO | wx.ICON_WARNING)
         confirmed = ret == wx.YES
         if confirmed:
-            # workspace_path should update after open dictionary/suite
+            # workspace_path should update after open directory/suite
             # There're two scenarios:
-            # 1. path is a dictionary
+            # 1. path is a directory
             # 2. path is a suite file
             new_path = RideFSWatcherHandler.get_workspace_new_path()
             if new_path and os.path.exists(new_path):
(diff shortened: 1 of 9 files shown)


src/robotide/ui/mainframe.py
```diff
@@ -593,6 +592,37 @@ def CreateSizeReportCtrl(self, width=80, height=80):
...
+               'Answering <No> will overwrite the changes on disk.']
+        if self._controller.is_dirty():
+            msg.insert(2, 'Answering <Yes> will discard unsaved changes.')
+        ret = wx.MessageBox('\n'.join(msg), 'Files Changed On Disk',
+                            style=wx.YES_NO | wx.ICON_WARNING)
+        confirmed = ret == wx.YES
+        if confirmed:
+            # workspace_path should update after open dictionary/suite
+            # There're two scenarios:
+            # 1. path is a dictionary
```
HelioGuilherme66, line 607: You mean "directory" not "dictionary" ;)

jnhyperion: yes, my typo.

PR #2247, opened 2020-07-23
Fix issue 2159

jnhyperion:
- Fixed some log message issue
- Fixed issue 2159
- Fixed a invalid cases selection issue

here's the reproduce step:

1. select one case, and run
2. go to text edit and apply some changes
3. the case will become unchecked
4. click run, you'll find that case is still selected


jnhyperion:
> I need to do more testing on the remove_invalid_cases_selection, specially when called, because it's not clear for me.

Did you reproduce this issue?

PR #2249, opened 2020-07-24
Optimize user experience

jnhyperion:


Nyral:
I agree with the removal of the double click logic. You can easily go to kw definition page by accident. 

Multi line edit doesn't work though and the way `OnKeyDown()` was modified is not OK. I'll add details below.

src/robotide/editor/kweditor.py
```diff
@@ -568,24 +567,23 @@ def OnKeyDown(self, event):
             if keycode == wx.WXK_SPACE:
                 self._open_cell_editor_with_content_assist()  # Mac CMD
             elif specialkcode in [wx.WXK_DOWN, wx.WXK_UP]:
-                self._skip_except_on_mac(event)
```
Nyral, line 571: put this back, move_rows doesn't work on windows

src/robotide/editor/kweditor.py
```diff
@@ -568,24 +567,23 @@ def OnKeyDown(self, event):
...
-            elif specialkcode in [wx.WXK_RETURN, wx.WXK_BACK]:
-                self.save()
-                self._move_grid_cursor(event, specialkcode)
+            elif specialkcode == wx.WXK_BACK:
+                self.OnDelete(event)
+                return
+            elif specialkcode == wx.WXK_RETURN:
+                if not self.IsCellEditControlShown():
+                    self._open_cell_editor()
+                    return
```
Nyral, line 583: I know you're trying to make it work like in a spreadsheet, but I think this way is more intuitive when editing RF tests.

For enter button:
- when not in edit mode, will enter edit mode of current cell
- when in edit mode, will exit edit mode and select cell to its right

For back button:
- select cell to the left
This is primarily used to go back to the cell on the left, after enter button is used. Deleting the content would be confusing for those used to work this way.


```
            elif specialkcode == wx.WXK_BACK:
                self._move_grid_cursor(event, specialkcode)
            elif specialkcode == wx.WXK_RETURN:
                if self.IsCellEditControlShown():
                    self._move_grid_cursor(event, specialkcode)
                else:
                    self._open_cell_editor()
```

src/robotide/lib/robot/writer/formatters.py
```diff
@@ -61,7 +61,7 @@ def _is_indented_table(self, table):
 
     def _escape_consecutive_whitespace(self, row):
         return [self._whitespace.sub(self._whitespace_escaper,
-                                     cell.replace('\n', ' ')) for cell in row]
+                                     cell.replace('\n', '${\\n}')) for cell in row]
```
Nyral, line 64: Why not just replace '\n' with '\\n'?

src/robotide/editor/kweditor.py
HelioGuilherme66: Nyral explained perfectly the current and desired behavior.

src/robotide/lib/robot/writer/formatters.py
HelioGuilherme66: This is not a required change, or what would be our benefit from it?
Usually we don't change the files under ../lib/robot they are a copy of RobotFramework, in this case v3.1.2.

On my fork, on branch `rf32` I am upgrading that library to v3.2.1. Which is really problematic. I created a `compat` directory to help making a transition between the old and new parser. I am doing a break, I have a minimal convertion working, it runs without errors but the object in the Tree are not OK, yet. I'll see if I can rebase to current development, and  I hope I can get your collaboration (@Nyral maybe we can talk about this). Anyway, this is not planned for v2.0, because it is taking too long.

Nyral: > Usually we don't change the files under ../lib/robot they are a copy of RobotFramework, in this case v3.1.2.
I think this improvement could be made without modifying any of the files from there.

> what would be our benefit from it?
I can see some benefit in making tests look better. Say for example that I want to log some data. This change will make it look exactly like in report.
image

src/robotide/editor/kweditor.py
jnhyperion: I only tested this PR in Mac, let me test more in windows and linux.

src/robotide/editor/kweditor.py
jnhyperion: - for the back button, I mix up the back & delete button, I'll revert it.
- for the enter button, @Nyral you suggestion is a better solution.
```
                if self.IsCellEditControlShown():
                    self._move_grid_cursor(event, specialkcode)
                else:
                    self._open_cell_editor()
```

src/robotide/lib/robot/writer/formatters.py
jnhyperion: > Why not just replace \n with \\n?

Actually I tried, did not work somehow.

>> Usually we don't change the files under ../lib/robot they are a copy of RobotFramework, in this case v3.1.2.

> I think this improvement could be made without modifying any of the files from there. Good point.

You're correct, let me find some better solution.

src/robotide/editor/kweditor.py
jnhyperion: @Nyral  tested in windows and linux, move row function works well, that line can be removed.

jnhyperion:
image
image
image

I tested grid cell multi lines feature, it works well.

src/robotide/lib/robot/writer/formatters.py
HelioGuilherme66: We would like to have visualization/navigation with hidden \n, but keep them on edit cell mode.

src/robotide/editor/kweditor.py
Nyral: Interesting... did you test in Windows 10 or 7?

src/robotide/editor/kweditor.py
```diff
@@ -568,24 +567,25 @@ def OnKeyDown(self, event):
             if keycode == wx.WXK_SPACE:
                 self._open_cell_editor_with_content_assist()  # Mac CMD
             elif specialkcode in [wx.WXK_DOWN, wx.WXK_UP]:
-                self._skip_except_on_mac(event)
                 self._move_rows(specialkcode)
             elif specialkcode == wx.WXK_RETURN:
-                self._move_cursor_down(event)
-            else:
-                event.Skip()
```
Nyral, line 576: Also rewrite this as follows, in order to keep both old and new behavior
```
            elif specialkcode == wx.WXK_RETURN:
                if self.IsCellEditControlShown():
                    event.GetEventObject().WriteText('\n')
                else:
                    self._move_cursor_down(event)
            else:
                event.Skip()
```

Nyral, line 576: Also rewrite this as follows, in order to keep both old and new behavior
```
            elif specialkcode == wx.WXK_RETURN:
                if self.IsCellEditControlShown():
                    event.GetEventObject().WriteText('\n')
                else:
                    self._move_cursor_down(event)
            else:
                event.Skip()
```

src/robotide/editor/kweditor.py
jnhyperion: yes, win7 & win10 with the combination of wxPython 4.0.7.post2 & wxPython 4.1.0

src/robotide/editor/kweditor.py
jnhyperion: Recovered old behaviour alt + enter -> move cursor down.

the last event.Skip() is not necessary is because, it eventually will go to the last line "event.Skip()" of method OnKeyDown()

Nyral: Lets leave it like this for now, I'll have to refactor that anyways after your merge as there are some shortcuts that don't need event.Skip().

src/robotide/editor/kweditor.py
Nyral: @HelioGuilherme66 Does it work for you?

Nyral: Ok. If it doesn't work then we can always find a fix for it in the future.

HelioGuilherme66:
I am asking users what is best for the newlines in Grid Editor: https://forum.robotframework.org/t/request-for-comments-new-lines-in-grid-editor/559

HelioGuilherme66:
@jnhyperion @Nyral Thank you for the PR and comments. I will merge, now. If the Poll results in changing the edit style, a later PR will be done (possibly to have it as an user option).

Issue #2255, opened 2020-07-26
Ctrl+Space causes entire column to be selected in grid


Nyral:
![image](https://user-images.githubusercontent.com/39315668/88490404-1315bf80-cfa4-11ea-924a-553cf2b0a8cb.png)

OS: Windows, Linux

HelioGuilherme66:
I think this is Low importance, because after selecting an item or cancelling the selection is cleared.

PR #2256, opened 2020-07-26
Removed event skip condition that was causing incorrect behavior

Nyral:
#2255

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -11,15 +11,23 @@ and this project adheres to http://semver.org/spec/v2.0.0.html[Semantic Versioni
 
 === Added
 
-- None
+- Added multiline view in Grid Editor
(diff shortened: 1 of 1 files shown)


src/robotide/editor/kweditor.py
```diff
@@ -584,8 +584,6 @@ def OnKeyDown(self, event):
                 self._open_cell_editor()
             else:
                 event.Skip()
-        if specialkcode != wx.WXK_RETURN:
```
HelioGuilherme66, line 587: If we remove this block, then in Linux, for keyboards that need AltGr to get { or [, the enclose text would not work. (AltGr-7 -> {, AltGr-8 -> [)

Nyral: Ok, I'll find a fix for that too after #2249 is merged.

PR #2257, opened 2020-07-26
Filemenu

HelioGuilherme66:


src/robotide/ui/mainframe.py
```diff
@@ -170,6 +171,7 @@ def __init__(self, application, controller):
         self._plugin_manager = PluginManager(self.notebook)
         self._review_dialog = None
         self._view_all_tags_dialog = None
+        self._curdir = None
```
jnhyperion, line 174: self._curdir means to save current external file dir right?

so a more specified name will be better, it probably will make us confused with current workspace dir

src/robotide/ui/mainframe.py
```diff
@@ -37,15 +37,16 @@
...
 
 _menudata = """
 [File]
 !&New Project | Create a new top level suite | Ctrlcmd-N
 ---
 !&Open Test Suite | Open file containing tests | Ctrlcmd-O | ART_FILE_OPEN
-!Open &Directory | Open directory containing datafiles | Shift-Ctrlcmd-O | \
-ART_FOLDER_OPEN
+!Open &Directory | Open directory containing datafiles | Shift-Ctrlcmd-O | ART_FOLDER_OPEN
+!Open External File | Open file in Code Editor | | ART_FOLDER_OPEN
```
jnhyperion, line 49: For now, "Open &Directory" & "Open External File" are using the exact same icon in menu bar, can we update to a new icon?

PR #2264, opened 2020-08-05
Adds RIDE as Application on MacOS

HelioGuilherme66:
Users will have to add same paths from $PATH environment variable to `/etc/paths` .
(Blame Apple for this)


src/bin/RIDE.app/Contents/MacOS/RIDE
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/Library/Frameworks/Python.framework/Versions/3.7/bin/ride.py $*
```
jnhyperion, line 2: we can update this file in postinstall _create_desktop_shortcut_mac method.

and then move this app pkg to ~/Desktop or directly to /Applications

HelioGuilherme66: Yes, I thought of that too. But no time to implement. Just was a proof of concept.
Most important is to solve the failure to find robot from shell. This may be because there is no shell window. We had that problem before on Windows. I would like to keep the current way to launch robot.

If we cannot manage it, then we can call robot from python.

HelioGuilherme66: More info:
- When RIDE.app is started from terminal with `open -a RIDE my_test_file.robot` still opens without that file, but with the latest edited.
- Robot will run when opened like this. So it really needs the shell.

jnhyperion: https://github.com/robotframework/RIDE/compare/MacOS_shortcut...jnhyperion:MacOS_shortcut?expand=1

I implemented this part based on your PR, you can take a look.

HelioGuilherme66: I did look, and it is working OK. But the dialog to create desktop shortcut appears, but does not create anything.
The environment variables are not set (I have them defined from .bash_profile.
When I change the run profile to use the full path to `robot` it manages to Run, but then it failed because chromedriver was not found.

jnhyperion: this branch copies RIDE.app to /Applications, you should find it in Mac Finder->Applications or from Launchpad.

HelioGuilherme66: Yes, I did find it. But the problem is that ignores the $PATH or /etc/paths. So it cannot find robot and others when need to run.

I was referring to the Shortcut creation, maybe we need to create a symlink to /Applications/RIDE.app

HelioGuilherme66: I think I found the solution here https://docs.python.org/3/using/mac.html#configuration

I also did not know that we can use `pythonw` on MacOS like we do in Windows.

jnhyperion: @HelioGuilherme66 this will work:

1. use `os.environ['PATH']` get os env variables
2. add this variable in to the app shell script like `export PATH="System variables"`

HelioGuilherme66: Finally I found a working solution (please confirm):
```
#!/bin/sh
PAL=$PATH
for i in `cat /etc/paths`
  do
     PAL="$PAL:$i"
  done
PATH=$PAL
export $PATH
# echo $PATH
/Library/Frameworks/Python.framework/Versions/3.7/bin/python3 -m robotide.__init__ $* 2> /dev/null &
```

HelioGuilherme66: @jnhyperion @Nyral We can test now the installation (and confirm it is running tests) with:
```
pip install -i https://test.pypi.org/simple/ robotframework-ride==2.0b2.dev1
```

jnhyperion: @HelioGuilherme66 there're 2 critical  known issues in current master branch:

https://github.com/robotframework/RIDE/pull/2272
https://github.com/robotframework/RIDE/pull/2273

we should merge them first.

jnhyperion: Confirmed, Mac OS shortcut feature works pretty well.

But the post install dialog hang issue still exists.

HelioGuilherme66: No it is not working the creation of /Applications/RIDE.app and desktop shortcut.
```
ride_postinstall.py 
Result True
  File "/Library/Frameworks/Python.framework/Versions/3.7/bin/ride_postinstall.py", line 26, in <module>
    postinstall.caller(None, sys.platform.lower())
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/robotide/postinstall/__main__.py", line 290, in caller
    return create_desktop_shortcut(platform, frame)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/robotide/postinstall/__main__.py", line 274, in create_desktop_shortcut
    return _create_desktop_shortcut_mac(frame)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/robotide/postinstall/__main__.py", line 222, in _create_desktop_shortcut_mac
    with open(app_script, 'w+') as shortcut:
FileNotFoundError: [Errno 2] No such file or directory: '/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/robotide/postinstall/RIDE.app/Contents/MacOS/RIDE'
```

PR #2267, opened 2020-08-09
Improve text editor

HelioGuilherme66:


HelioGuilherme66:
I have tested this in MacOS and the Del is not working correctly. Only works when text is selected, and deleted chars are less one than it should be.
There are differences in keyboard actions, depending if Grid Editor is active or not.

PR #2279, opened 2020-08-18
Add app icon in dock & window size fix

jnhyperion:
1. Added RIDE icon in MacOS dock.
2. Fix frame size is incorrect when maximize app window
close / maximum /minimum button is hidden due to the incorrect window size, see the screenshot
image


HelioGuilherme66:
Nice one.
Thanks @jnhyperion.

PR #2289, opened 2020-08-24
Fixed mac os post install args parsing issue

jnhyperion:


src/robotide/postinstall/__main__.py
```diff
@@ -211,29 +211,28 @@ def _create_desktop_shortcut_linux(frame=None):
...
+            shortcut.write("#!/bin/sh\nPAL=$PATH\nfor i in `cat /etc/paths`\n    do\n        PAL=\"$PAL:$i\"\n"
+                           "    done\nPATH=$PAL\nexport $PATH\n{} -m robotide.__init__ $* 2>"
+                           " /dev/null &\n".format(sys.executable))
+        if exists(ride_app_pc_path):
+            shutil.rmtree(ride_app_pc_path, True)
         shutil.copytree(ride_app_module_path, ride_app_pc_path)
-    user = str(subprocess.check_output(['logname']).strip(), encoding='utf-8')
-    userdesktoplink = '/Users/' + user + '/Desktop/' + ride_app_name
-    if not exists(userdesktoplink) or option_f:
+        user_desktop_link = os.path.join(os.path.expanduser('~'), 'Desktop', ride_app_name)
```
HelioGuilherme66, line 230: @jnhyperion This code should not be changed, This is the second time you do this.
I have coded this way because when the user is installing with `sudo -H` my code returns the username. With your code it will return `root`.

HelioGuilherme66 (changes requested):
Other than renaming variables, the user desktop detection should stay as it was.

src/robotide/postinstall/__main__.py
jnhyperion: @HelioGuilherme66 it's because in my Mac the result is contrary to your results:

`
print(str(subprocess.check_output(['logname']).strip(), encoding='utf-8'))
print(os.path.expanduser('~'))

>> root
>> /Users/jnhuang
`

jnhyperion: In my PyCharm terminal: `logname -> root`
In my system termimal: `logname -> jnhuang`

I think we should not use system command to get user desktop path, because it has system configuration dependencies, while os.path module not.

If user use 'sudo -H' install RIDE, which means user intent to install RIDE under user root.

HelioGuilherme66: I think if the user uses `sudo -H` it means install for all users, for example if is an IT people installing remotely in the network.

HelioGuilherme66 (approved):

PR #2296, opened 2020-08-28
Publisher upgrade

jnhyperion:
Upgrade publisher module, using pubsub instead of self-implemented publisher.

this will fix object will be held forever if listener is subscribe but not unsubscribed in the end of the object life circle.

and will fix many potential memory leak issue.


src/robotide/publish/messages.py
```diff
@@ -64,16 +70,8 @@ def publish(self):
...
-            self._publish(self)
-        except Exception as err:
-            self._publish(RideLogException(
-                message='Error in publishing message: ' + str(err),
-                exception=err, level='ERROR'))
-            raise # DEBUG on Python3 we have a loop
-
-    def _publish(self, msg):
-        publisher.PUBLISHER.publish(msg.topic, msg)
+        from robotide.publish.publisher import PUBLISHER
```
Nyral, line 73: Please don't do imports inside of the method.

src/robotide/ui/preview.py
```diff
@@ -63,14 +63,14 @@ def OnShowPreview(self, event):
...
-            self._panel.tree_node_selected(event.item)
+            self._panel.tree_node_selected(message.item)
 
-    def OnTabChanged(self, event):
-        self._update_preview()
+    def OnTabChanged(self, message):
+        self._update_preview(None)
 
-    def _update_preview(self, event=None):
+    def _update_preview(self, message):
```
Nyral, line 73: In `OnShowPreview` this method is called without any arguments. It will cause error.

src/robotide/publish/messages.py
HelioGuilherme66: Sometimes internal importing is the only solution to solve circular imports.

jnhyperion: Yes, because there's a circular import if we put the import in the file head.

src/robotide/ui/preview.py
jnhyperion: Updated

src/robotide/publish/messages.py
Nyral: ok

src/robotide/log/log.py
```diff
@@ -82,7 +82,8 @@ def _create_menu(self):
         self.register_action(ActionInfo(
             'Tools', 'View RIDE Log', self.OnViewLog, position=84))
 
-    def _log_message(self, log_event):
+    def _log_message(self, message):
+        log_event = message
```
Nyral, line 86: Kinda redundant. Why not just replace ``log_event`` with ``message`` everywhere in the method?

src/robotide/editor/settingeditors.py
```diff
@@ -190,7 +192,10 @@ def OnClear(self, event):
         self._controller.execute(ClearSetting())
         self._update_and_notify()
 
-    def update_value(self, event=None):
+    def _ps_on_update_value(self, message):
+        self.update_value()
+
+    def update_value(self):
```
Nyral, line 198: Why create a separate method instead of writing just message=None ?

jnhyperion: there're 2 reasons:

- the method arg `event` is never used in all project
- the overwritten method in class `DocumentationEditor` also does not have the arg, their args signature should be the same

src/robotide/log/log.py
jnhyperion: I thought it could be a little confusing if replace all `log_event` to `message`.

But we should avoid redundant, updated.

HelioGuilherme66 (approved):

Issue #2290, opened 2020-08-25
Resource file will disappear when saving in text editor


jnhyperion:
1. Open resource file in text editor
2. make some changes and save
3. save 3 times
this resource file will disappear, only happens in windows

PR #2297, opened 2020-08-31
Fix resource import path issue

jnhyperion:
- Fixed resource file will disappear after saving from Text Editor
- Fixed duplicated resource file/folder in tree nodes

[TestCases.zip](https://github.com/robotframework/RIDE/files/5149027/TestCases.zip)


jnhyperion:
will fix: https://github.com/robotframework/RIDE/issues/2290

HelioGuilherme66 (approved):

PR #2306, opened 2020-09-02
FindText returns a tuple if something is found...

ImmoWetzel:
see ticket


ImmoWetzel:
See also .. https://docs.wxpython.org/wx.stc.StyledTextCtrl.html#wx.stc.StyledTextCtrl.FindText

jnhyperion (changes requested):
You should update this method: _find_text_position

PR #2309, opened 2020-09-02
search_new_text_issue_#2308

ImmoWetzel:


jnhyperion (changes requested):
Please do not combine 2 PRs, this issue also happens in wxPython 4.0.7.post2, in this version, there's not tuple return value issue, you can use this version to fix this bug.

PR #2310, opened 2020-09-03
Implemented "Sort Variables" and "Sort Tests" functionalities in tree menu

Nyral:
image

In addition to that, I also added Move Up and Move Down to variable tree menu.

image


Nyral:
Test units and CHANGELOG updated

Issue #2317, opened 2020-09-13
The new PUBLISHER changes, now shows a Pop-up when failed to subscribe a topic.


HelioGuilherme66:
1 - Open RIDE and disable the Editor (Grid Editor) on Manage Plugins
2- Keep the Text Editor plugin active.
3- Close RIDE.
4- Open RIDE
(a pop-up dialog will appear with failure to subscribe ChangeTab, and blocks the normal start of RIDE)

Must be changed to silently continue to load RIDE, or to log in RIDE Log, or stdout print.

PR #2318, opened 2020-09-14
Make publisher errors slient

jnhyperion:
- LogPlugin should be loaded in the first, because some error message could be printed in log plugin during other plugin loading
- Title in plugin manage setting page cannot show properly

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -45,7 +45,7 @@ and this project adheres to http://semver.org/spec/v2.0.0.html[Semantic Versioni
 
 === Fixed
 
-- Fixed incorrect tittle in manage plugin settings
+- Fixed incorrect title in manage plugin settings
 - Fixed search in Text Editor with wxPython 4.1.0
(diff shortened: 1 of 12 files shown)


CHANGELOG.adoc
```diff
@@ -45,6 +45,7 @@ and this project adheres to http://semver.org/spec/v2.0.0.html[Semantic Versioni
 
 === Fixed
 
+- Fixed incorrect tittle in manage plugin settings
```
HelioGuilherme66, line 48: tittle? :D

jnhyperion:
#2317

jnhyperion:
this PR will cause unit test print lots of error message, need to avoid

CHANGELOG.adoc
jnhyperion: updated

jnhyperion:
fixed error msg in unit

HelioGuilherme66 (approved):

PR #2341, opened 2020-11-20
Improve "Run" tab

dpantelin:
1) Increased width of left part window on launching app
2) Improved style of "Run" tab
3) Improved of building arguments for run RobotFramework
4) Added possibility of saving console log to file


HelioGuilherme66 (changes requested):
I requested changes on a PR.

HelioGuilherme66 (approved):

utest/contrib/testrunner/test_command_builder.py
```diff
@@ -0,0 +1,56 @@
+#  Copyright 2008-2015 Nokia Networks
```
HelioGuilherme66, line 1: A totally new file should have copyright starting in 2020.

utest/contrib/testrunner/test_command_builder.py
```diff
@@ -0,0 +1,56 @@
...
+#  limitations under the License.
+
+import unittest
+
+from robotide.contrib.testrunner.CommandBuilder import CommandBuilder
+
+
+class CommandBuilderStub(CommandBuilder):
+    def _get_listener_path(self):
+        return 'C:\\My Work\\Python\\TestRunnerAgent.py'
```
HelioGuilherme66, line 23: This cannot be dependent on the operating system, at least not Windows, but POSIX.

utest/contrib/testrunner/test_command_builder.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_build_command_default(self):
+        builder = CommandBuilderStub()
+        result = builder.build()
+        self.assertEqual(result, '')
+
+    def test_build_command_full(self):
+        builder = CommandBuilderStub()
+        builder.set_prefix('prefix')
+        builder.add_arg_file('C:\\User name\\Temp\\Ride\\arg_file.txt')
```
HelioGuilherme66, line 36: We should use our unit tests resources directory, and relative paths. Using forward slashes should work in all oprating systems.

HelioGuilherme66 (approved):
Only important changes in Unit tests. I will make them and merge.

utest/contrib/testrunner/test_command_builder.py
HelioGuilherme66: OK, let's keep it because is only text. I was worried if real paths.

HelioGuilherme66 (approved):

utest/contrib/testrunner/test_command_builder.py
dpantelin: I changed "builder". FileWriter is static class. It's not good, I hade some problems with testing my logic and decided move FileWriter to another place. This paths for example only. To be honest, I don't know how path look like on Linux. We can add here a few additional tests for Linux.

dpantelin: And this is my style to writing unit tests. I usually provide some common examples how it works. I prefer to use the most realistic values for variables and names in tests. I hope it will be usefull for people, who will try to understand how is it working.

Issue #2342, opened 2020-11-24
Python library with keyword only arguments fails to load


JFoederer:
Both Python and Robot framework have had updates to support named arguments, but when such a library is loaded in RIDE it fails. More specifically it fails when any of the named arguments also includes a default value.

Take for instance this basic example
`def kw(self, *, arg1=None):`

Which results in this error in the RIDE log:
`Traceback (most recent call last):
list index out of range
  File "C:\Python37\lib\site-packages\robotide\spec\librarymanager.py", line 87, in _fetch_keywords
    return get_import_result(path, library_args)
  File "C:\Python37\lib\site-packages\robotide\spec\libraryfetcher.py", line 29, in get_import_result
    ) for kw in lib.handlers]
  File "C:\Python37\lib\site-packages\robotide\spec\libraryfetcher.py", line 29, in <listcomp>
    ) for kw in lib.handlers]
  File "C:\Python37\lib\site-packages\robotide\spec\libraryfetcher.py", line 40, in _parse_args
    parsed[index] = parsed[index] + '=' + str(args.defaults[value])  # DEBUG str(value)`

HelioGuilherme66:
@JFoederer Thank you for the report and PR.

I would like to know what version of RIDE you are using: released==1.7.4.2 Beta==2.0b1 or from `master` (I assume this is the one). Also the version installed of Robot Framework (RIDE uses internally, the 3.1.2 and some code is copied from it).

About the problem. You say:

> Take for instance this basic example
> ```def kw(self, *, arg1=None):```

I believe that this is invalid code for Robot Framework standards, because the varargs should come in last position.
So the code should be (but I did not confirmed):
```def kw(self, arg1=None, *):```

Did you find an example library with your case?
Thanks.

JFoederer:
Hi @HelioGuilherme66,

Thank you for the quick response. After running into the issue I have first upgraded to 2.0b1 and I have taken the current master as a base for my changes. My installed Robot framework version is 3.2.2 on a Windows machine.

The example was encountered in a proprietary library at a client's site. I first suspected the library, but checking against the Robot documentation revealed that the implementation was actually correct, supported and had a purpose.
https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#getting-keyword-arguments

The feature is called _named-only arguments_ The difference is in how you can call the keyword with arguments

`def kw(self, arg1=None, *args):`

allows for both positional and named usage:
`kw    value`
`kw    arg1=value`

[...]

PR #2343, opened 2020-11-24
Include kwonlyargs support

JFoederer:
Fix the issue (#2342) where python keyword libraries fail to load in RIDE due to named arguments. e.g.:
def kw(self, *, arg1=None):

### DIFF
--- a/src/robotide/spec/libraryfetcher.py
+++ b/src/robotide/spec/libraryfetcher.py
@@ -32,14 +32,12 @@ def get_import_result(path, args):
 
 def _parse_args(args):
     parsed = []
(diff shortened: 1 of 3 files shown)


JFoederer:
Good to see the tests being run here as well, although these failures don't seem to be related to the changes.

I had a hard time myself as well to get RIDE 2.0b1 installed on Ubuntu. Also due to wxpython issues. I only succeeded after manually installing all dependencies and then using `pip --no-deps`. Haven't gotten the run tab working yet though.

HelioGuilherme66 (approved):

PR #2344, opened 2020-11-25
Improved run of tests

dpantelin:
1) Modified creation of command arguments - used fluent builder
2) Modified creation of command for RobotFramework - used fluent builder
3) Modified run of tests:
  - refactoring of _run_tests method
4) Updated unit tests for creating command
5) Added unit tests for creating command arguments


src/robotide/contrib/testrunner/FileWriter.py
```diff
@@ -41,6 +42,8 @@ class FileWriter:
 
     @staticmethod
     def write(file_path, lines, windows_mode, mode='w'):
+        if not os.path.exists(os.path.dirname(file_path)):
+            os.makedirs(os.path.dirname(file_path))
         if IS_WINDOWS:
```
jnhyperion, line 47: in non-Windows platform, this method will raise exception, because `lines` could be bytes

jnhyperion (changes requested):

PR #2352, opened 2020-12-17
Keyword auto suggestion in grid editor without shortcut

jnhyperion:
Works properly in wxPython 4.0.7.post2 and 4.1.1

There will be some issue in wxPython 4.1.0, due to wxPython bug: https://github.com/wxWidgets/Phoenix/issues/1797

And luckily, this PR also fixed a tricky crash issue: `Fixed RIDE will crash when using third party input method in Mac OS`, although I don't know the exact root cause.


HelioGuilherme66 (changes requested):
This looks a good idea.
I tried in my local machine, and I am not totally satisfied with the behavior.
- Only worked at first time, next cell edits dir not show list,
- Ctrl-Space did not call the list
- When typing and the list present, the cell contains all kw text from found entry.
  I would prefer to only have the text we are typing and update the list.

HelioGuilherme66:
@jnhyperion Thank you for this PR. Sorry for taking almost 3 years to complete.

I need to close this, because too many changes and conflicts are resolved in #2577

PR #2382, opened 2021-04-06
Fix --loglevel option handling

JFoederer:
Fix for the failure when using the command line option for loglevel, e.g. --loglevel DEBUG:INFO
```
File ".\Python38\lib\site-packages\robotide\contrib\testrunner\TestRunnerAgent.py", line 260, in _send_socket
    self.filehandler.flush()
ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine
```

When the first log message arrives it compares the loglevel with the level that is set. Comparison is done as number, but the internal variable was directly stored as its string value causing an exception in the comparison.

The fix is to use the already available translation table from loglevels to numeric values. Note that checking the validity of the argument is already done by Robot, so this code is not reached when invalid values are entered. This prevents the need for further error handling.


HelioGuilherme66 (commented):
Excellent Pull Rquest!

HelioGuilherme66 (commented):
Excellent Pull Request :)

HelioGuilherme66:
This is going to be reverted, because it had side effects :(

PR #2494, opened 2022-10-09
fix bug #2361

zhujunling-nj:
Chinese name of test case can not display correct on console log

PR #2502, opened 2022-11-06
Fix for diacritic characters display on RIDE console

Harm10:
This is only for windows platform and when encoding settings are different (often the case in Western Europe)


HelioGuilherme66 (changes requested):
@Harm10 Thanks for your effort, however I will prefer the PR #2494 because it is simpler and from the source of content.

If you want, you can remove the code change, and leave only the typo correction, then I will merge this doc fix.

HelioGuilherme66:
@Harm10 You did well.
I hope you can write more PR.
(but next time you need to create a branch and work on it. See the CONTRIBUTING document )

Issue #2512, opened 2022-12-29
Make all Keyboard shortcuts common between Grid and Text Editors.


HelioGuilherme66:
For example:
[x] Ctrl-1 to 5 were missing on Text Editor. 
[x] Ctrl-T is missing in Grid Editor.
[x] Ctrl-D, Ctrl-I are missing in Text Editor.

Harm10:
I first concentrated on getting the same options in the Grid Editor and Edit pull down menu. I also rearranged their sequences to get the same on both. I also looked at the menu for row number select.

I have some trouble in understanding the key registrations in the texteditor. Ctrl-T is active there but where is this set and where is the code? I cannot find it in texteditor.py
And what is Ctrl-T compared to Alt-up and Alt-down? Those Alt codes are not active in texteditor. To me Ctrl-T seems to be superfluous and should be replaced by Alt-up like used by the grid editor. Swap a row is the same as move the current row up?

Also already the report that shortcut Alt-space for Content Assist does not work properly in the grid editor (Ctrl-space does work properly).
I also find sometimes 3 different key bindings for this.
So the question first is: which key shortcuts should start Content Assist?
Perhaps handling all Content Assist logic should go into another issue?

The texteditor has shortcut Del commented out. Is this with reason?

HelioGuilherme66:
>Ctrl-T is active there but where is this set and where is the code? I cannot find it in texteditor.py
We do not set it. It is by default from scintilla (stc) library. For example, just try it on Notepad++ :)
I'd like to have it in Grid Editor, just to be equal to Text Editor.

There is a small difference on Ctrl-T  versus Alt-Up, the cursor remains in the same position. In Grid Editor, Alt-Up, keeps the whole moved row highlighted.
Of course that Alt-Up/Down are priority key actions to implement in Text Editor.

Alt-Space works on Text Editor but not in Grid Editor. Text should be corrected to Ctrl-Alt-Space for Grid Editor. Text is correct for Tools->Content Assist.
(this combination existed in the past, but can not be used in Windows and Mac)

There was a time that Del was needed in Windows. Also I did try to implement in Mac, but that key does not exist in Mac keyboards.

Harm10:
OK. Will try to add Ctrl-T in grid editor, row display menu and Edit menu. Also Alt-up/down in text editor.

Btw Is there a way to influence the right click menu on Text editor? It only displays the default (Windows?) settings currently I feel.

[...]

PR #2513, opened 2023-01-11
Grid updates for #2512

Harm10:
This update is only related to shortcuts for the grid editor. That is the Edit pull down menu, the row right click menu and the cell right click menu. The help shortcuts information has been changed as well.
Text editor will follow in another PR.

I left the Content Assist shortcuts to be picked up by me later.
Also I noticed that the Help information sometimes mentions Shift-Ctrl instead of Ctrl-Shift which is in my view the way you say that. Or is there another reason for this difference in help text?

I have added:
Undo / Redo / Move cursor down / Swap Row Up and also changed some text in the menu. Also the order within the different menus is now the same if the same shortcut is active in that menu.

### DIFF
--- a/src/robotide/context/__init__.py
+++ b/src/robotide/context/__init__.py
@@ -152,8 +152,8 @@ SHORTCUT_KEYS = '''\
         <td>Stop running test suite</td>
     </tr>
 </table>
-<h3>Grid</h3>
 
+<h3>Grid</h3>
 <table>
     <tr align="left">
         <th><b>Shortcut</b></th>
@@ -172,20 +172,8 @@ SHORTCUT_KEYS = '''\
         <td>Pop-up JSON Editor</td>
     </tr>
     <tr>
-        <td>CtrlCmd-I</td>
-        <td>Insert row(s)</td>
-    </tr>
-    <tr>
-        <td>CtrlCmd-D</td>
-        <td>Remove row(s)</td>
-    </tr>
-    <tr>
-        <td>Shift-CtrlCmd-I</td>
-        <td>Insert cell(s)</td>
-    </tr>
-    <tr>
-        <td>Shift-CtrlCmd-D</td>
-        <td>Remove cell(s)</td>
+        <td>CtrlCmd-B</td>
+        <td>Go to Definition</td>
     </tr>
     <tr>
         <td>CtrlCmd-Z</td>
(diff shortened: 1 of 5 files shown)


src/robotide/context/__init__.py
```diff
@@ -248,11 +260,11 @@ def bind_keys_to_evt_menu(target, actions):
         <td>Paste (does not move cells or rows)</td>
     </tr>
     <tr>
-        <td>Shift-CtrlCmd-V</td>
+        <td>CtrlCmd-Shift-V</td>
         <td>Insert (adds empty rows and pastes data)</td>
     </tr>
     <tr>
-        <td>Delete</td>
+        <td>Del</td>
```
HelioGuilherme66, line 267: Del is an abbreviation, why not using Delete?

src/robotide/editor/__init__.py
```diff
@@ -25,24 +25,37 @@
...
 Insert Cells | Insert Cells | Shift-Ctrl-I
 Delete Cells | Delete Cells | Shift-Ctrl-D
 Insert Rows | Insert Rows | Ctrlcmd-I
 Delete Rows | Delete Rows | Ctrlcmd-D
+---
+Select All | Select All | Ctrlcmd-A
+---
+Cu&t | Cut | Ctrlcmd-X
+&Copy | Copy | Ctrlcmd-C
+&Paste | Paste | Ctrlcmd-V
```
HelioGuilherme66, line 55: These functions are now in the bottom of a long list of functions. They should have stayed on the top.

src/robotide/editor/macroeditors.py
```diff
@@ -51,6 +51,9 @@ def close(self):
     def save(self):
         self.kweditor.save()
 
+    def movecursordown(self):
```
HelioGuilherme66, line 54: This is redundant, of move_cursor_down (and see comment there).

src/robotide/editor/macroeditors.py
```diff
@@ -85,17 +91,38 @@ def delete_rows(self):
...
 
-    def comment(self):
+    def comment_rows(self):
         self.kweditor.OnCommentRows()
 
-    def uncomment(self):
+    def uncomment_rows(self):
         self.kweditor.OnUncommentRows()
 
+    def move_cursor_down(self):
```
HelioGuilherme66, line 100: This function here makes a duplicate call. It was working without this here,
Same will happen with the create variables function.

src/robotide/editor/macroeditors.py
```diff
@@ -85,17 +91,38 @@ def delete_rows(self):
...
+
     def comment_cells(self):
-        self.kweditor.OnCommentCell(None)
+        self.kweditor.OnCommentCells(None)
 
     def uncomment_cells(self):
-        self.kweditor.OnUncommentCell(None)
+        self.kweditor.OnUncommentCells(None)
+
+    def make_variable(self):
```
HelioGuilherme66, line 118: These functions repeat the action. For example, you'll get ${${}} .

HelioGuilherme66 (commented):
Apart for the repeated actions, and the long context menus, this is acceptable. However, I'll create a new PR with changes, so I can release 2.0b3 today.

Thanks, Harm!

HelioGuilherme66:
This PR is completed by #2515

src/robotide/context/__init__.py
Harm10: On the grid cell menu the description of the shortcut is also Del. I followed the same syntax, same order to get the same collection.

src/robotide/editor/__init__.py
Harm10: They are also at the bottom of the grid cell menu. Again following the rule: same content, same order

src/robotide/editor/macroeditors.py
Harm10: OK

src/robotide/editor/macroeditors.py
Harm10: I think you added it in singular but I am not sure.

Harm10:
About your comment "the long context menus": that is what you get if you have a lot of shortcuts and want to show them all.

Harm10:
As you made some changes I should wait on the new release and then clone again?

Issue #2464, opened 2022-07-14
Performance drop loading test suite


JFoederer:
I have been running some tests on RIDE version 2.0b2.dev11 and noted a significant performance drop when loading test suites. Opening a test suite consisting of approximately 2500 robot files previously took just seconds to load, now it takes more than 5 minutes.

The amount of keyword libraries also plays a role. There are some variations, but the majority uses a predefined set of approximately 50 keyword libraries, varying in size from a few to one hundred keywords each.

I am running on a Windows 10 machine, Intel i7 with 16GB RAM, no other load on the system. Python 3.8.7 and Robot framework 5.0.1.

I haven't been able to keep up with recent updates, so I can't really give an estimate on when the performance regression was introduced.

HelioGuilherme66:
@JFoederer I have some ideas where the problem may be. I would like for to try to generate sample data, with similar conditions of your test suites. For that, we have a script that you can call. Here is some help:
```
invoke generate-big-project --help
Usage: inv[oke] [--core-opts] generate-big-project [--options] [other tasks here ...]

Docstring:
  Generate big test data project to help perf testing.

Options:
  -a STRING, --args=STRING
  -i, --install
  -u, --upgrade


C:\GitHub\RIDE>invoke generate-big-project
rfgen: Running under Python 3
Test project is created into directory (option 'd'): .\tmp\testdir
Generating test project with following settings
    5 test libraries (option 'l')
    10 keywords per test library (option 'k')
    1 test suites (option 's')
    10 tests per test suite (option 't')
    3 test steps per test case (option 'e')
    1 resource files (option 'f')
    0 external resource files (option 'g')
    30 resources per resource file (option 'r')

    Seed was 1657788693
    
```
I you create one, please post here the command you used (preferably with a  fixed  random seed). then we can benchmark with it.
Thanks.

JFoederer:
That looks like a very useful tool!
What is the syntax for passing the options? It keeps telling me it doesn't know any of the options.

HelioGuilherme66:
Sorry, I did tried today but seems that it is not working. I suspect changes on argsparser module.

axiom41:
Hi all... Maybe i'm wrong but as far as i could see, the performance degradation when loading the test suites is related to the indentation problems (issue #2460), because RIDE (starting from 2.0b2.dev11) is adding the extra spaces since the suite loading, even without making any change. The more test suites/test cases you have, the more you have to wait for RIDE to add such spaces.
[...]

Issue #2423, opened 2021-11-03
RIDE freezes on every action when large number of variables are imported from file


amitlohar:
In out project, we import variables from 3 different files. Total count of these variables is around **2900**.
We have a main master resource file in which we declare what all is to be imported.

When we try to create a testcase, each and every action takes ~10secs to respond for example ctrl+s action, or even adding/writing something in cell and pressing TAB takes ~10secs to complete/respond. This freeze time increase as the number of lines in test case grow.

The workaround we do for this is:
1. Once the RIDE is opened/loaded, we remove these 3 files import line.
2. Reload (close and open) RIDE
3. Then add the removed lines from Step1, (do not reload RIDE). After this things are normal until RIDE again reloaded with those import lines present.

RIDE ver = v2.0b1 running on Python 3.7.7 Same issue happens even on stable release of RIDE.
robotframework  = 4.1.2
robotframework-seleniumlibrary =5.1.3
Windows Server 2016, 16gb RAM 2cpu

chevi:
We have imports from file and slow performance of the Ride too...
Will test this issue, after fix

amitlohar:
Tested on the new beta build, (2.0b3), issue still exists.
Python 3.7.7
robotframework = 6.0
robotframework-seleniumlibrary =6.0.0
Windows Server 2016, 16gb RAM 2cpu

HelioGuilherme66:
There are improvements that may solve your problem, even without code changes in RIDE.
If the variables are defined in a *.robot file (imported as a Resource), those files should be renamed to *.resource.
If the files with variables definitions are in a normal directory separated from tests, try to rename that directory with a prefix `_`.
So, "resources" or "lib" would be "_resources" and "_lib". With this, Robot Framework and RIDE will not look for test suites there.

amitlohar:
So we have all our keywords written in Resources folder with .robot extension. I will try to change it to .resource.
Adding _ prefix does not show the folder in tree view which is downside for me, since i need them to read/edit if changes.
Hopefully changing extension to .resource will help.

I will try and test it out soon.

HelioGuilherme66:
@amitlohar 
>Adding _ prefix does not show the folder in tree view which is downside for me, since i need them to read/edit if changes.
This should not happen. The folders will be visible in RIDE, only they will not be searched for tests. If resource files are not used by any test suite, they will appear in gray color.

(please use the current development version from `master` if you can. Thanks.)

amitlohar:
[...]

Issue #1683, opened 2017-02-13
Performance Enhancement While opening project with few thousand files


vinaybond:
When I open my project in RIDE, it take around 5 minutes to open it. There are around 3-4 thousand .robot file available in my project. This loads much faster in eclipse. Can the performance be improved?

Issue #1602, opened 2016-02-19
pool performance when loading many resource/case files from a suite


divfor:
I have a suite that have 100+ case/resource .txt files. when I open RIDE it spends 4s to load data files and 45s to populate from these data files. PyCharm's profilor reports that there are thread lock conflicts that lead to too much wait time.
![diagram](https://cloud.githubusercontent.com/assets/3319979/13164273/f31365f0-d6ee-11e5-992c-ca2a3b7aec20.png)

divfor:
def _load_datafile(self, path, load_observer):
        datafile = self._loader.load_datafile(path, load_observer) #spend 4s
        if not datafile:
            return None
        self._populate_from_datafile(path, datafile, load_observer) #spend 36s
        return datafile

divfor:
def resources_for(self, datafile, load_observer):
        return self._load(_ResourceLoader(
            datafile, self._namespace.**get_resources**), load_observer)

divfor:
I addressed the pool performance is located in below three functions. can we improve it?
```python
 def _collect_each_res_import(self, datafile, ctx, collector):
        items = set()
        ctx.set_variables_from_datafile_variable_table(datafile)
        for imp in self._collect_import_of_type(datafile, robotapi.Resource):
            res = self._resource_factory.get_resource_from_import(imp, ctx)
            if res and res not in ctx.parsed:
                ctx.parsed.add(res)
                collector(res, ctx, items)
        return items

    def get_resources_from(self, datafile):
        resources = list(self._get_resources_recursive(datafile,
                                                       RetrieverContext()))
        resources.sort(key=operator.attrgetter('name'))
        return resources

    def _get_resources_recursive(self, datafile, ctx):
        resources = set()
        res = self._collect_each_res_import(datafile, ctx, self._add_resource)
        resources.update(res)
        for child in datafile.children:
            resources.update(self.get_resources_from(child))
        return resources
```

yanne:
I am not sure how easy it is to optimise that code, but it's definitely worth taking a look.

kolesnikandrei:
I think it needs to move on to asynchronous actions

PR #2514, opened 2023-01-14
Improves Performance when loading big projects

HelioGuilherme66:
Reduces cells representation with as_list() by using cells[] in Step creation.


HelioGuilherme66:
Probably solves #1602 #1683 #2423 #2464

PR #2528, opened 2023-01-26
Ctrl-Shift-I - insert cell(s)

Harm10:
It turned out to be rather complicated. Wordleft is great for shifting to the left but does not see the tab between cells as a word. So I coded 3 extra functions to get what I needed: _get_cell_no, _get_number_of_cells and _get_position_of_cell
I definitely need those functions for the next command Ctrl-Shift-D (Delete cell(s)).
The amount of cells to insert and the place are all decided by the selection in the first line. If there are more lines selected the conclusion made on the first line is based on its cell number.

So I will first await your review on this one. I tried to add some comments on what is happening. Looking at the other code this does not seem to be the habit to do (why not?).
There are also 2 TODO's in this code: one about the undo of a multiple insert and one about positioning not working on a line that starts with an empty cell.
I hope it will also work on your machine (and not break other things :-) )


src/robotide/editor/texteditor.py
```diff
@@ -1056,6 +1057,105 @@ def execute_uncomment(self, event):
...
+            if cellpos > findpos:
+                cell_no -= 1
+                break
+        return cell_no
+
+    def _get_number_of_cells(self, begpos, endpos):
+        # get number of cells in range begpos-endpos
+        # Warning! GetStringSelection does not work properly if there are diacritics in the content above (!) the selected range
+        the_content = self._editor.GetTextRange(begpos, endpos)
+        celltab = ' ' * self._tab_size
```
JFoederer, line 1128: Will this also work when working on files that have a different tab size, but haven't been modified yet?

I ask because I have seen some issues where I had the tab size set to 4, but then when I pulled files from the archive that were 2-space separated, it was interpreted as a single cell.

Harm10: On positioning the current tab size is taken into account as set by _tab_size. Or do you mean something else?
There is a lot of code working with this _tab_size.

On another note: I did not get any mail or so on your review. Should I have gotten some notice?

PR #2711, opened 2024-02-08
Fix issues with uninstalled locales

JFoederer:


JFoederer:
I was still having issues on the system that did not have broad locale support. After a lot of searching and trial and error I arrived at this solution that does not change anything directly on Python's locale, but still seems to support all translations. System dialog buttons are translated when the language is available, but stay in the default language if the locale is not installed.

For certain languages it already fails from wxpython, but I was unable to check if wxpython can also have partial installs with some languages missing. In that case I get a dialog when switching, or at startup. This is still better than the crash we used to get.
image
After closing the dialog one can still reach the preferences dialog to set a supported language.

HelioGuilherme66 (approved):
I included your PR in another,

Issue #2719, opened 2024-02-14
FileExplorer enabled setting is never set to False


JFoederer:
By default the File Explorer is opened, which is also stored in the `settings.cfg` file. Unless this file edited manually there is no way to persist the _opened_ setting at value False. I would expect that if the File Explorer is closed using the cross on the panel or by toggling it from the _View_ menu, that this choice is remembered.

HelioGuilherme66:
The File Explorer is a Plugin, and the Enable state stored in `settings.cfg` is the Plugin state. If we Disable the Plugin, then the Menu option is removed.

The toggle idea, to Show/Hide the File Explorer (and maybe the Project Explorer) and persist this state, is a good idea.

PR #2720, opened 2024-02-14
Make View File Explorer a toggle option

JFoederer:
Fix for issue #2719


JFoederer:
This is not a complete fix yet. For some reason I am unable to bind to the close event of the panel to the object. For now, the close button is removed and the menu option (F11) is changed into toggle functionality. This enabled the intended behaviour, but I think the cross button on the panel is the preferred way.

Issue #2703, opened 2024-01-30
Supporting Robot Framework 7's new VAR syntax


JFoederer:
In Robot Framework 7 a [new syntax for declaring variables](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#var-syntax) is introduced. To complete RF7 support from RIDE this new syntax should be supported. Support would entail syntax highlighting and adding declared variables to the suggestions while editing.

HelioGuilherme66:
Partially done in v2.1dev53.
Still missing to fix multiline variables in Variables section.
Still missing to get names/values for VAR defined variables.

HelioGuilherme66:
Missing variables recognition in autocomplete :(.

PR #2813, opened 2024-06-23
Add support for VAR marker

HelioGuilherme66:
Still missing to fix multiline variables in Variables section.
Still missing to get values for VAR defined variables.


HelioGuilherme66:
Closes #2703

PR #2916, opened 2024-12-12
Add Chinese language support for 'zh'

xianglongfei-8888:
Fixed the issue of not being able to create desktop shortcut icons, successfully tested on Deepin and UOS Linux operating systems

image
image
image

### DIFF
--- a/src/robotide/postinstall/__init__.py
+++ b/src/robotide/postinstall/__init__.py
@@ -178,7 +178,7 @@ def _create_desktop_shortcut_linux(frame=None):
     # DEBUG: Add more languages
     desktop = {"de": "Desktop", "en": "Desktop", "es": "Escritorio",
                "fi": r"Työpöytä", "fr": "Bureau", "it": "Scrivania",
-               "pt": r"Área de Trabalho"}
+               "pt": r"Área de Trabalho", "zh": "Desktop"}
     user = getlogin()
     try:
         ndesktop = desktop[DEFAULT_LANGUAGE[0][:2]]


HelioGuilherme66:
@xianglongfei-8888 Thanks for this Pull Request.
I would like to talk to you about the translation of RIDE to zh_CN. Is there a way we can talk in Instant Chat (our Slack, #ride-dev), or other way?

xianglongfei-8888:
@HelioGuilherme66 Thank you for your reply.
I currently do not have a strong demand for an English interface that can meet daily use, but being able to support it would be the best. Does the RIDE tool interface need to be translated into Chinese?

HelioGuilherme66:
>Does the RIDE tool interface need to be translated into Chinese?
It already supports, you can change it on Tools->Preferences->General =>Language.
We have the translation files in our source code, but we synchronize with the https://crowdin.com/project/robotframework-ride
. 
Best is to have people confirming the quality of the translation.

Other thing is the actual test cases that can have the *Language:* setting, then the settings labels appear translated on the GUI.

xianglongfei-8888:
@HelioGuilherme66
That's great, I have checked the language settings of RIDE and it supports Chinese translation.

If I find any translation issues, I will submit them.

xianglongfei-8888:
@HelioGuilherme66 
What's the reason? It can't pass. Can you help me check? 
Thank you！

HelioGuilherme66:
@xianglongfei-8888 

>What's the reason? It can't pass. Can you help me check?

It is a problem known, only happens on Python 3.13 on Fedora 41 tests. It is not related with your changes,

PR #3018, opened 2026-03-29
Optimize grid expansion performance

chelslava:
## Summary
Optimizes the grid expansion logic in `GridEditor` to improve performance when editing large test files.

## Problem
The original `_expand_if_necessary` method used `while` loops to add rows/columns one at a time:
```python
while self.NumberRows <= max(1, row+1, 10-row):
    self.AppendRows(1)  # Called multiple times
```

Each `AppendRows()`/`AppendCols()` call triggers:
- UI update events
- Grid recalculation
- Potential screen refresh

For large files or rapid editing, this could cause noticeable lag.

## Solution
Calculate the needed expansion once and add in batch:
```python
needed_rows = max(1, row + 2) - self.NumberRows
if needed_rows > 0:
    self.AppendRows(needed_rows)  # Single call
```

## Changes
- Replaced while loops with batch expansion
- Removed commented-out debug code
- Simplified the logic for clarity

## Performance Impact
- Reduces number of UI update events
- Improves responsiveness when editing large test suites
- No functional changes to user-facing behavior

## Testing
Tested by editing test files with 100+ rows and columns - no visual or functional differences observed, but editing feels snappier.


chelslava:

PR #3019, opened 2026-03-29
Fix wildcard imports and resource leak

chelslava:
## Summary
This PR fixes wildcard imports and a resource leak in the RIDE codebase.

### application/__init__.py
- Replaced `from .debugconsole import *` with `from . import debugconsole`
- The module is used as `debugconsole.start(ride)` in `__init__.py`, so the wildcard import was unnecessary

### publish/__init__.py
- Added comment explaining why the wildcard import is acceptable (messages.py properly defines `__all__`)
- Fixed resource leak in `get_html_message()`:
  - Before: `open(...).read()` - file handle never closed
  - After: `with open(...) as f: return f.read()` - proper context manager
- Modernized string formatting to use f-string
- Added explicit `encoding='utf-8'` for cross-platform compatibility

## Testing
- Python syntax validation passed
- The changes maintain backward compatibility


chelslava:

PR #3025, opened 2026-03-29
Add visible searching indicator to Find Usages dialog

### LINKED ISSUE #2672: Move 'searching...' indicator to more visible location
When using the '_Find usages_' feature in RIDE, the search can take quite some time. More than once have I missed results, or wrongfully assumed there were no hits, due to the way the dialog indicates the search is still ongoing.

Currently the title of the dialog is used to display a 'searching...' text, but typically a user would be looking at the results table in the dialog.
image

A simple fix could be to reuse the 'searching...' text as indicator, but instead put it as the next row in the table. I am sure there a many other ways as well.

_submitted on RIDE v2.0.8.1_

chelslava:
## Summary

- Adds a "Searching..." row as the first row in the Find Usages dialog table during search
- Makes the search progress indicator more visible to users who focus on the results table
- The indicator row disappears automatically when the search completes

Closes #2672

### DIFF
--- a/src/robotide/usages/usagesdialog.py
+++ b/src/robotide/usages/usagesdialog.py
@@ -30,6 +30,7 @@ class UsagesDialog(RIDEDialog):
         self._dots = None
         self._name = name
         self._selection_listeners = []
+        self._searching = False
         title = "'%s'" % name
         RIDEDialog.__init__(self, title=title, size=(650, 400))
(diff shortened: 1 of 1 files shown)


HelioGuilherme66:
I have merged this PR in my fork, and will close this when my `develop` branch is merged-

PR #3031, opened 2026-03-30
Add periodic auto-save feature

### LINKED ISSUE #1376: Automatic save
> *Originally submitted to [Google Code](http://code.google.com/p/robotframework-ride/issues/detail?id=1336) by guillaum... on 4 Mar 2014*

It could be a good idea to have an automatic save of all file modified, every minute for exemple. Period could be configured.

chelslava:
## Summary
Closes #1376

Adds a periodic auto-save feature that automatically saves all modified files at a configurable interval.

## Changes
- Added auto-save interval setting to preferences
- Added auto-save timer in mainframe
- Default interval: 0 (disabled)

### DIFF
--- a/src/robotide/ui/mainframe.py
+++ b/src/robotide/ui/mainframe.py
@@ -380,7 +380,7 @@ class RideFrame(wx.Frame):
 
     def _on_auto_save(self, event):
         """Auto-save all files when timer fires."""
-        _ = event
+        __ = event 
         if self.controller and self.controller.is_dirty():
             RideBeforeSaving().publish()
             self.save_all()
--- a/src/robotide/ui/mainframe.py
+++ b/src/robotide/ui/mainframe.py
@@ -167,7 +167,9 @@ class RideFrame(wx.Frame):
         self.color_foreground = self.general_settings.get('foreground', '#5E5C64') if self.general_settings else '#5E5C64'
         self.font_face = self.general_settings.get('font face', '') if self.general_settings else ''
         self.font_size = self.general_settings.get('font size', 11) if self.general_settings else 11
-        self.ui_language = self.general_settings.get('ui language', 'English') if self.general_settings else 'English' 
+        self.ui_language = self.general_settings.get('ui language', 'English') if self.general_settings else 'English'
+        self._auto_save_interval = application.settings.get('auto save interval', 0)
+        self._auto_save_timer = None
         self.main_menu = None
         self._init_ui()
         self.SetIcon(wx.Icon(self._image_provider.RIDE_ICON))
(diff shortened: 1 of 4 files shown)


src/robotide/ui/mainframe.py
```diff
@@ -364,6 +368,30 @@ def get_selected_datafile(self):
...
+            self._auto_save_timer.Stop()
+            self._auto_save_timer = None
+        if self._auto_save_interval > 0:
+            self._auto_save_timer = wx.Timer(self)
+            self.Bind(wx.EVT_TIMER, self._on_auto_save, self._auto_save_timer)
+            self._auto_save_timer.Start(self._auto_save_interval * 60 * 1000)  # minutes to milliseconds
+
+    def _on_auto_save(self, event):
+        """Auto-save all files when timer fires."""
+        _ = event
```
HelioGuilherme66: To avoid breaking translation function,  when we use it, we should use:
__ = event

HelioGuilherme66 (changes requested):
See inline comment.

src/robotide/ui/mainframe.py
HelioGuilherme66: Please do a change and new commit. It must be double underscore to not change the original function.
```
__ = event  # Not _ = event 
```

HelioGuilherme66 (approved):

src/robotide/ui/mainframe.py
```diff
@@ -364,6 +368,30 @@ def get_selected_datafile(self):
...
+            self.Bind(wx.EVT_TIMER, self._on_auto_save, self._auto_save_timer)
+            self._auto_save_timer.Start(self._auto_save_interval * 60 * 1000)  # minutes to milliseconds
+
+    def _on_auto_save(self, event):
+        """Auto-save all files when timer fires."""
+        __ = event 
+        if self.controller and self.controller.is_dirty():
+            RideBeforeSaving().publish()
+            self.save_all()
+            self.SetStatusText(_('Auto-saved all files'))
```
HelioGuilherme66, line 387: I think it is missing the restart of the timer. I propose:
```
wx.CallAfter(self._start_auto_save_timer)
```

HelioGuilherme66 (changes requested):
Please add the reset of the timer.

src/robotide/ui/mainframe.py
chelslava: Done, thanks. I added `wx.CallAfter(self._start_auto_save_timer)` after the successful auto-save path and pushed the update in commit `e44f0fd5`.

I also added a focused regression test for this behavior.

HelioGuilherme66 (approved):

PR #3055, opened 2026-07-17
fix: recognize variable-width space separators

### LINKED ISSUE #3045: Multi-space separators not respected
Given RIDE is configured to use 4 separating spaces
and there is a file on disk that uses 2 separating spaces
When RIDE loads the file
Then all arguments are merged into the keyword text

Example:
```
*** Test cases ***
First test case
    Keyword with two arguments  arg1  arg2
```

Loads as: `Keyword with two arguments arg1 arg2`
and after editing the file causes: `FAIL : No keyword with name 'Keyword with two arguments arg1 arg2' found.`

w3lld1:
## Summary

- I updated separator detection to recognize Robot Framework's two-or-more-space separators even when RIDE is configured to write four spaces.
- I kept tab separators supported and added a regression test for the reported four-space configuration with a two-space input file.

## Testing

- `xvfb-run -a python -m pytest -q utest/lib/robot/parsing/test_robotreader.py` — 1 passed
- `python -m py_compile src/robotide/lib/robot/parsing/robotreader.py utest/lib/robot/parsing/test_robotreader.py`
- `git diff --check origin/develop...HEAD`

I also attempted the full `invoke test` suite locally. Its first application pass completed with 15 passed and 3 skipped, but the main run reported failures in `utest/controller/test_cellinfo.py` and `utest/application/test_postinstaller.py` before timing out while running `utest/application/test_updatenotifier.py`.

Fixes #3045

### DIFF
--- a/src/robotide/lib/robot/parsing/robotreader.py
+++ b/src/robotide/lib/robot/parsing/robotreader.py
@@ -170,23 +170,11 @@ class RobotReader(object):
                 return
             if not self._cell_section:
                 return
-            spc = idx = nospc = 0
-            for idx in range(0, len(line)):
-                if line[idx] != ' ':
-                    nospc += 1
-                    # if spc <= 2:
-                    #     spc = -1
-                    #
-                    if spc >= 2:
-                        break
-                    spc = 0
-                elif line[idx] == ' ':  # and nospc > 0:
-                    spc += 1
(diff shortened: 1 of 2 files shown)


HelioGuilherme66:
@w3lld1 Your changes have been incorporated in #3059, because needed some improvements to not break some test suites editing.
Thanks for the PR.

Issue #3063, opened 2026-08-07
Unable open the existing directories, when clicking om Open Directory option application is showing any console Wizard


Laxman11r:
As a QA engineer, I installed the Robotframework setup and RIDE editor.
Next I created one directory and one project. If I am trying to open the existing directory RIDE is not opening any console wizard

~~Failed to upload "RIDE Issue Directory not opening.mp4"~~

Note: The original content meant to have a screen recording of RIDE. After JFoederer comment, the problem was identified and a fix was created.

The problem appears after a first time run of RIDE, when the settings.cfg does not have the default_dir correctly set. Other problem could be the wxPython new version, 4.3.0.

JFoederer:
I fully agree with the `bad_issue_report` tag, but there might actually be something to report here. I was installing a fresh pc earlier this week; Windows with Python 3.14. After installing RIDE, simply the latest release, I could not use '_Open Directory_'. The button did not do anything. In other words, the '_Select folder dialog_' did not open.

At that time I only needed a single file, so I opened it directly using '_Open Test Suite_'. I haven't gotten back to that pc since, but if I do and I find something, I will drop it here.

HelioGuilherme66:
@JFoederer I see no problems with the current development version, on Windows 10 , Python 3.13.5 and wxPython 4.2.5.

It may be that the problem only happens on the very first time RIDE is run, right after installation.

JFoederer:
Yes, sorry. I forgot to mention that I also upgraded my regular pc and that it still worked fine there.

This is the exception I get on the fresh pc:

> Traceback (most recent call last):
  File "C:\Users\me\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\robotide\ui\actiontriggers.py", line 366, in __call__
    action.act(event)
    ~~~~~~~~~~^^^^^^^
  File "C:\Users\me\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\robotide\action\action.py", line 89, in act
    self.action(event)
    ~~~~~~~~~~~^^^^^^^
  File "C:\Users\me\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\robotide\ui\mainframe.py", line 673, in on_open_directory
    path = wx.DirSelector(message=_("Choose a directory containing Robot files"),
                          default_path=self.controller.default_dir)
TypeError: DirSelector(): 'default_path' is not a valid keyword argument

HelioGuilherme66:
[...]

PR #3064, opened 2026-08-08
fix open dir for wxpython 4.3

JFoederer:
Resolution for issue #3063

Even though the documentation states that the argument is correct: https://docs.wxpython.org/wx.functions.html#wx.DirSelector

It also states that the arguments are the same meaning as those of [DirDialog.__init__](https://docs.wxpython.org/wx.DirDialog.html#wx.DirDialog.__init__). Only there they changed: `default_path` -> `defaultPath`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ Project: https://github.com/robotframework/RIDE#readme
 long_description_content_type = "text/x-rst"
 license = "Apache-2.0"
 license-files = { paths = ["LICENSE.txt"] }
-requires-python = ">=3.10, <3.15"
+requires-python = ">=3.9, <3.15"
 authors = [
     { name = "Robot Framework Developers", email = "robotframework@gmail.com" },
 ]
@@ -35,6 +35,7 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "License :: OSI Approved :: Apache Software License",
     "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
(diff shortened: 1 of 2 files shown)


JFoederer:
Pipelines fail, because there are a lot of version-specific wxPython dependencies in the jobs.
@HelioGuilherme66, what is your backwards compatibility strategy?

src/robotide/ui/mainframe.py
```diff
@@ -671,7 +671,7 @@ def on_open_directory(self, event):
         __ = event
         if self.check_unsaved_modifications():
             path = wx.DirSelector(message=_("Choose a directory containing Robot files"),
-                                  default_path=self.controller.default_dir)
+                                  defaultPath=self.controller.default_dir)
```
HelioGuilherme66: We must condition to wxPython version.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ Project: https://github.com/robotframework/RIDE#readme
 long_description_content_type = "text/x-rst"
 license = "Apache-2.0"
 license-files = { paths = ["LICENSE.txt"] }
-requires-python = ">=3.9, <3.15"
+requires-python = ">=3.10, <3.15"
```
HelioGuilherme66: I think it is a bit too soon to drop 3.9. Maybe when 3.15 is out.

pyproject.toml
```diff
@@ -49,11 +48,10 @@ dependencies = [
     "Pygments>=2.20.0",
     "PyPubSub",
     "Pywin32; platform_system == 'Windows'",
-    "wxPython",
+    "wxPython>=4.3.0",
```
HelioGuilherme66: Too soon to change. In my Fedora 42, would need to compile because no package available.

pyproject.toml
```diff
@@ -109,8 +107,8 @@ readme = "README.adoc"
 packages = [{include = "robotide"}]
 
 [tool.poetry.dependencies]
-python = "^3.9"
-wxPython = "^4.1.1"
+python = "^3.10"
+wxPython = "^4.3.0"
```
HelioGuilherme66: See previous comments.

HelioGuilherme66 (changes requested):
Please condition to wxPython versions, only change mainframe.py.

src/robotide/ui/mainframe.py
JFoederer: That was indeed the alternative and answers my question. Update pending.

HelioGuilherme66 (approved):

PR #3065, opened 2026-08-08
Fixed broken directory dialog on first RIDE execution after install

HelioGuilherme66:
Fixes #3063


JFoederer (commented):
Looks good when combined with the pull request including the backwards compatible fix. Thanks for doing the admin!

